PR #6, opened 2021-06-04
bugfix: fix error of using --prefer-interfaces for multiple interfaces

### LINKED ISSUE #5: use --prefer-interfaces to assign an interface list failed
#### What happened:
There are two types of nodes, some of them use eth1 as node container networking nic, some of them use eth0 as node container networking nic

I wish to use configuration of "--prefer-interfaces=eth0,eth1" with only one rama-daemon ds to support all of the nodes, but some thing failed, pod is stuck at Containercreating period with `ensure vlan interface  error: Link not found`

#### What you expected to happen:
Pod can be created correctly on all of the nodes.

#### How to reproduce it (as minimally and precisely as possible):
Using configuration of "--prefer-interfaces=eth0,eth1" to deploy a rama-daemon ds, and need an environment of different nic name on different nodes

#### Anything else we need to know?:
NONE

#### Environment:
- rama version: unreleased version
- OS (e.g: `cat /etc/os-release`): CentOS
- Kernel (e.g. `uname -a`): 4.19
- Kubernetes version: 1.16
- Install tools: trident
- Others:

hhyasdf:

1. Fix --prefer-interfaces with interface list error
2. Remove build binaries.

Fixes #5  
### Describe how you did it

Reassign config.NodeVxlanIfName after finding a valid prefer interface.

### Describe how to verify it

Logs rama-daemon and will print out the single correct prefer interface link `use ethxxx as node vlan interface` while using a prefer interface list.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/config/config.go
+++ b/pkg/daemon/config/config.go
@@ -104,7 +104,7 @@ func ParseFlags() (*Configuration, error) {
 	// mute info log for ipset lib
 	logrus.SetLevel(logrus.WarnLevel)
 
-	flag.Set("alsologtostderr", "true")
+	_ = flag.Set("alsologtostderr", "true")
 	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
 	klog.InitFlags(klogFlags)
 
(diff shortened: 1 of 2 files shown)


mars1024 (approved):
/lgtm
/approve

PR #13, opened 2021-06-09
remove overlay logs for underlay-only schema

hhyasdf:
Pull Request Description
---

#### Describe what this PR does / why we need it

Fix an error that rama-daemon pod seems print a lot of "vtep information error", which will always exists in underlay-only schema. 

#### Does this pull request fix one issue?

NONE

#### Describe how you did it

Remove overlay logs if only underlay network exist.

#### Describe how to verify it

Deploy an underlay environment and look through the logs of rama-daemon.

#### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -425,56 +425,58 @@ func (c *Controller) iptablesSyncLoop() {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


mars1024 (approved):
/lgtm

PR #19, opened 2021-06-23
feature: print runtime git commit

hhyasdf:
Pull Request Description
---

Print runtime commit id of Rama in rama-manager and rama-daemon logs. It helps us to find the correct commit id while debugging. 

NONE

### Describe how you did it
Link a flag "main.gitVersion" while building binaries, and add logs in rama-daemon and rama-manager.

### Describe how to verify it
Run rama-daemon and rama-manager pod, logs of "starting rama manager with git version: <commit id>" will be printed at the first place. Executing command of "cat /rama/COMMIT_ID" in pod will show commit id too.

### Special notes for reviews

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -11,10 +11,12 @@ RUN export GOCACHE=/tmp && \
     export GOARCH=${ARCH} && \
     export CGO_ENABLED=0 && \
     export GOOS=linux && \
-	go build -o dist/images/rama -ldflags "-w -s" -v ./cmd/cni && \
-	go build -o dist/images/rama-daemon -ldflags "-w -s" -v ./cmd/daemon && \
-    go build -o dist/images/rama-manager -v ./cmd/manager && \
-    go build -o dist/images/rama-webhook -v ./cmd/webhook
+    export COMMIT_ID=`git rev-parse --short HEAD 2>/dev/null` && \
+    go build -o dist/images/rama -ldflags "-w -s" -v ./cmd/cni && \
+    go build -ldflags "-w -s -X \"main.gitCommit=`echo $COMMIT_ID`\" " -o dist/images/rama-daemon -v ./cmd/daemon && \
+    go build -ldflags "-X \"main.gitCommit=`git rev-parse --short HEAD 2>/dev/null`\" " -o dist/images/rama-manager -v ./cmd/manager && \
+    go build -o dist/images/rama-webhook -v ./cmd/webhook && \
+    echo $COMMIT_ID > ./COMMIT_ID
 
 RUN cd /go/src/github.com/oecp/rama/dist/secrets && \
     sh generate-tls-certificates.sh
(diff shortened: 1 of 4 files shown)


mars1024 (changes requested):
IMO, I don't think this will help us a lot, because
1. log won't be kept for a long time, it rotates usually
2. commit id is not useful compared with tag, which could be found in image url

hhyasdf:
> IMO, I don't think this will help us a lot, because
> 
> 1. log won't be kept for a long time, it rotates usually
> 2. commit id is not useful compared with tag, which could be found in image url

For the first problem, I add an /rama/VERSION file into the image for long time usage.

For the second problem, image tag sometimes is unreliable. This happens mostly in develop environment, developer sometime don't know the exact code version if they dose not use a official release or have changed the image's tag.

Dockerfile.amd64
```diff
@@ -11,10 +11,12 @@ RUN export GOCACHE=/tmp && \
     export GOARCH=${ARCH} && \
     export CGO_ENABLED=0 && \
     export GOOS=linux && \
+    export GIT_VERSION=`git rev-parse --short HEAD 2>/dev/null` && \
 	go build -o dist/images/rama -ldflags "-w -s" -v ./cmd/cni && \
-	go build -o dist/images/rama-daemon -ldflags "-w -s" -v ./cmd/daemon && \
-    go build -o dist/images/rama-manager -v ./cmd/manager && \
-    go build -o dist/images/rama-webhook -v ./cmd/webhook
+	go build -ldflags "-w -s -X \"main.gitVersion=`echo $GIT_VERSION`\" " -o dist/images/rama-daemon -v ./cmd/daemon && \
+    go build -ldflags "-X \"main.gitVersion=`git rev-parse --short HEAD 2>/dev/null`\" " -o dist/images/rama-manager -v ./cmd/manager && \
```
mars1024, line 17: pay attention to the indent

Dockerfile.amd64
```diff
@@ -11,10 +11,12 @@ RUN export GOCACHE=/tmp && \
...
+    go build -o dist/images/rama-webhook -v ./cmd/webhook && \
+    echo $GIT_VERSION > ./VERSION
```
mars1024, line 19: VERSION seems puzzled to me, could we use another name?

cmd/daemon/daemon.go
```diff
@@ -30,10 +30,14 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 )
 
+var gitVersion string
+
 func main() {
 	klog.InitFlags(nil)
 	defer klog.Flush()
 
+	klog.Infof("Starting rama manager with git version: %v", gitVersion)
```
mars1024, line 39: nit: manager --> daemon

Dockerfile.amd64
hhyasdf: fixed

Dockerfile.amd64
hhyasdf: change to COMMIT_ID

cmd/daemon/daemon.go
hhyasdf: fixed

mars1024 (approved):
/lgtm
/approve
thanks!

PR #20, opened 2021-06-23
feat: use special annotation to reallocate/retain IP for stateful pods

mars1024:

Pull Request Description
---

Sometimes IP-retain on stateful pods will lead to other problems, for example, users lose control of pod scheduling after pod was created once, and even resources have been run out, pod will still be limited to these unscheduled nodes.

Introduce a new annotation switch to trigger IP reallocation for stateful pods.
Also bring in a flag `default-ip-retain` to set the global default choice.

### Describe how you did it
If pod have the reallocate annotation, it will be scheduled freely and allocated IPs will be recycled.

### Describe how to verify it
Annotate the pod template and watch IP change.

### Special notes for reviews
NONE


hhyasdf (approved):
It seems to be a small problem that, for the stateful pods which close the ip-retain flag, their ipinstances will not be released until their owner stateful workload be deleted or they are recreated themselves. 

But I think it can be ignored yet.

PR #22, opened 2021-06-25
bugfix: keep IP consistent when reallocating

mars1024:
Pull Request Description
---

keep IP consistent when reallocating

NONE

### Describe how you did it
When recycling IP, deleting IP instance instead of removing IP from IPAM.
Ignore terminating IP instance when listing.

### Describe how to verify it
NONE

### Special notes for reviews
NONE

### DIFF
--- a/pkg/webhook/validating/pod.go
+++ b/pkg/webhook/validating/pod.go
@@ -85,8 +85,8 @@ func PodCreateValidation(ctx context.Context, req *admission.Request, handler *H
 			if ipInstance.DeletionTimestamp == nil && ipInstance.Spec.Network != specifiedNetwork {
 				return admission.Denied(fmt.Sprintf(
 					"pod has assigned ip %s of network %s, cannot assign to another network %s",
-					ipList.Items[0].Spec.Address.IP,
-					ipList.Items[0].Spec.Network,
+					ipInstance.Spec.Address.IP,
+					ipInstance.Spec.Network,
 					specifiedNetwork,
 				))
 			}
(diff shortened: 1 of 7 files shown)


pkg/webhook/validating/pod.go
```diff
@@ -78,13 +78,18 @@ func PodCreateValidation(ctx context.Context, req *admission.Request, handler *H
...
 		}
-		if len(ipList.Items) >= 1 && ipList.Items[0].Spec.Network != specifiedNetwork {
-			return admission.Denied(fmt.Sprintf(
-				"pod has assigned ip %s of network %s, cannot assign to another network %s",
-				ipList.Items[0].Spec.Address.IP,
-				ipList.Items[0].Spec.Network,
-				specifiedNetwork,
-			))
+
+		for i := range ipList.Items {
```
hhyasdf, line 82: It's better to use "for _, ipInstance := range ipList.Items" instead of index only?

pkg/webhook/validating/pod.go
```diff
@@ -78,13 +78,18 @@ func PodCreateValidation(ctx context.Context, req *admission.Request, handler *H
...
+			var ipInstance = &ipList.Items[i]
+			// terminating ipInstance should be ignored
+			if ipInstance.DeletionTimestamp == nil && ipInstance.Spec.Network != specifiedNetwork {
+				return admission.Denied(fmt.Sprintf(
+					"pod has assigned ip %s of network %s, cannot assign to another network %s",
+					ipList.Items[0].Spec.Address.IP,
```
hhyasdf, line 88: ipInstance.Spec.Address.IP instead of ipList.Items[0].Spec.Address.IP?

pkg/webhook/validating/pod.go
mars1024: this will avoid deepcopy when ranging because element in `Items` is value type not pointer type

pkg/webhook/validating/pod.go
mars1024: good catch!

hhyasdf (approved):
/lgtm

PR #27, opened 2021-07-07
Nil interfaceToIPSliceMap when CreateNeighManager

Louhwz:
Pull Request Description
---

bug fix. 

None
### Describe how to verify it
no need
### Special notes for reviews
* Why this bug didn't happen before?
  * Because every time when calling `AddPodInfo` which would cause panic because of the bug,  `ResetInfos` will be called.


hhyasdf (approved):
/lgtm

PR #28, opened 2021-07-09
improvement: node local vxlan ip can be choosed

hhyasdf:
Pull Request Description
---

If node in a cluster has multiple ip addresses, or one of the ip address is the same for every node, rama overlay may work incorrectly.

### Describe how you did it
Only use vtep ip and address of node object as node local vxlan ip by default. Other ip addresses of node can be choose by "--extra-node-local-vxlan-ip-cidrs" parameter of rama-daemon.

### Describe how to verify it
Update rama-daemon in a exist cluster.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/node.go
+++ b/pkg/daemon/controller/node.go
@@ -73,7 +73,7 @@ func (nic *NodeIPCache) UpdateNodeIPs(nodeList []*v1.Node, localNodeName string)
 			return fmt.Errorf("parse node vtep mac %v failed: %v", node.Annotations[constants.AnnotationNodeVtepMac], err)
 		}
 
-		ipStringList := strings.Split(node.Annotations[constants.AnnotationNodeIPList], ",")
+		ipStringList := strings.Split(node.Annotations[constants.AnnotationNodeLocalVxlanIPList], ",")
 		for _, ipString := range ipStringList {
 			nic.nodeIPMap[ipString] = macAddr
 		}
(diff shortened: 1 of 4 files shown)


pkg/daemon/controller/node.go
```diff
@@ -228,10 +228,34 @@ func (c *Controller) reconcileNodeInfo() error {
...
+			}
+		}
+
+		if isNodeObjectAddr {
+			nodeLocalVxlanAddr = append(nodeLocalVxlanAddr, addr)
+			continue
+		}
+
+		// Add extra node local vxlan ip.
+		for _, cidr := range c.config.ExtraNodeLocalVxlanIPCidrs {
```
mars1024, line 248: Two questions,
1. why include ip but not exclude it? IMO, some invalid ip addresses should be ignored.
2. why multiple IP addresses will be pick up for vxlan devices, vtep seems to need only one?

hhyasdf: IP addresses for the vxlan devices are not only used to forward packets, which make node as a vtep gateway, but also to be used as the source addresses when node need to access pod.

mars1024: I don't think this is a good feature for us now, if pod want to access node, only the "main" IP will and should be chosen, even the node may have several IPs, IMO, these should be treated as external IPs.

mars1024: BTW, do you have some user cases about this?

hhyasdf: > I don't think this is a good feature for us now, if pod want to access node, only the "main" IP will and should be chosen, even the node may have several IPs, IMO, these should be treated as external IPs.

This is exactly this pr are doing. Before this pr, every node will get all of local ip addresses on it to assigned to vxlan interface. After this pr, only the "main" IP will and should be chosen.

> BTW, do you have some user cases about this?

Choosing extra node local ip addresses as local vxlan address is for extra usages, in case someone want to use it.

mars1024 (approved):
/lgtm
/approve

pkg/daemon/controller/node.go
mars1024: Gotcha!

PR #31, opened 2021-07-19
improvement: arp request source enhance

hhyasdf:
Pull Request Description
---

For rama underlay network, the packets of pods will get through host network without tunnel encapsulation. And for our implementation, node is actually the next-hop gateway for pod, which needs to do arp resolving for pod network.

If a node doesn't have an ip address in pod subnet, it will use the ip of the node itself as arp packet's sender ip while resolving pod gateway's address, which might cause problem if the host network will check the sender's ip and regards the arp packet invalid if the sender ip is not in the same subnet with target ip.

This pr will add maintain a local pod address on node vlan forward interface for every rama underlay subnet, the local pod address will be took as arp sender address by kernel.

NONE

### Describe how you did it
Maintain a local pod address on node vlan forward interface for every rama underlay subnet, change these addresses while ip instance is changed.

### Describe how to verify it
Need an environment with arp sender ip check.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/addr/addr.go
+++ b/pkg/daemon/addr/addr.go
@@ -24,6 +24,7 @@ import (
 	ramav1 "github.com/oecp/rama/pkg/apis/networking/v1"
 	"github.com/oecp/rama/pkg/daemon/containernetwork"
 
+	"github.com/containernetworking/plugins/pkg/ip"
 	"github.com/oecp/rama/pkg/constants"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
@@ -75,6 +76,8 @@ func (m *Manager) SyncAddresses(getIPInstanceByAddress func(net.IP) (*ramav1.IPI
 	}
 
 	existEnhancedAddrMap := map[string]map[string]netlink.Addr{}
+	existManualAddrSubnetMap := map[string]map[string]bool{}
+	existLinkMap := map[string]netlink.Link{}
 
 	for _, link := range linkList {
 		// ignore container network virtual interfaces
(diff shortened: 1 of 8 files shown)


mars1024 (commented):
IMO, I think there is no need to scan all addresses locally, we can use a force-sync at start-up then diff and re-sync in memory.
Make code logic more clear, maybe.

And I think we still need a global switch for this feature, cause turn this on will make data plane rely on control plane, and this should be chosen by our users.

pkg/daemon/addr/utils.go
```diff
@@ -0,0 +1,87 @@
...
+		LinkIndex: link.Attrs().Index,
+		Src:       addr.IP,
+	}, netlink.RT_FILTER_TABLE|netlink.RT_FILTER_OIF|netlink.RT_FILTER_SRC)
+
+	if err != nil {
+		return false, fmt.Errorf("list local routes for interface %v and src %v failed: %v",
+			link.Attrs().Name, addr.IP.String(), err)
+	}
+
+	if len(routeList) == 0 && addr.Flags&unix.IFA_F_SECONDARY == 0 {
```
mars1024, line 23: If enhanced address is what we need for the source IP of ARP request, I think we miss something here.

mars1024: If we are using the bridge mode, then the original IP of same subnet can be used for ARP request, but it won't hit this judgment.

pkg/daemon/addr/utils.go
```diff
@@ -0,0 +1,87 @@
+package addr
```
mars1024, line 1: Missing copyright.

pkg/daemon/addr/utils.go
hhyasdf: fixed~

pkg/daemon/addr/utils.go
hhyasdf: fixed~

pkg/daemon/addr/addr.go
```diff
@@ -0,0 +1,214 @@
...
+					existManualAddrSubnetMap[linkName] = map[string]bool{}
+				}
+				existManualAddrSubnetMap[linkName][cidr.String()] = true
+			}
+		}
+	}
+
+	// clear enhanced addresses which are impossible to be used
+	for existLinkName, existSubnetMap := range existEnhancedAddrMap {
+		existLink, err := netlink.LinkByName(existLinkName)
```
mars1024, line 122: This link is just listed before, so no need to check it existence, right?

hhyasdf: fixed~

mars1024 (approved):
/lgtm
/approve

PR #33, opened 2021-07-22
improvement: remove duplicated overlay subnet route entries when autoNatOutgoing is true

SwimilTylers:
Pull Request Description
---

remove duplicated overlay subnet route entries

NONE

### Describe how you did it

When overlay subnet route entries is added to toOverlaySubnet route table, DO NOT append it to allValidSubnet again.

### Describe how to verify it

check by `ip rule` and `ip route show table XXXX`

### Special notes for reviews

### DIFF
--- a/pkg/daemon/route/route.go
+++ b/pkg/daemon/route/route.go
@@ -224,7 +224,7 @@ func (m *Manager) SyncRoutes() error {
 		}
 	}
 
-	var allValidSubnet []*net.IPNet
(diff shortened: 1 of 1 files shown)


hhyasdf (commented):
Maybe variable name also should also be changed~

SwimilTylers:
> Maybe variable name also should also be changed~

Gotta

mars1024 (approved):
/lgtm, thanks!

hhyasdf (approved):

PR #36, opened 2021-07-28
bugfix: add routes for excluded ips

### LINKED ISSUE #30: Bad path from overlay pod to gateway of underlay subnets
Bug Report
---

Type: *bug report*

### What happened
Bad network path from overlay pod to gateway of underlay subnets, when ping underlay gateway from overlay pod, unexpected neighbor event comes from vxlan interface.

We expect overlay pod will react underlay gateway on a underlay+snat network path.

### How to reproduce it (as minimally and precisely as possible)
Ping underlay gateway from overlay pod.

### Anything else we need to know?
Beside underlay gateway, if we only use part of a underlay subnet, the rest IPs will have the same problem as gateway.

### Environment
- rama version:
- Kernel (e.g. `uname -a`):
- Kubernetes version:
- Install tools:
- Others:

hhyasdf:
Pull Request Description
---

Rama support a flexible ip range configuration, in which even a discontinuous ip address range can be described. While if a subnet CIDR has some excluded ip addresses, there would be some problem for overlay pod to access these excluded ip addresses.

Fixes #30 

### Describe how you did it
Four cases should be considered:

1. Underlay pod to access underlay excluded ip addresses: the same with underlay pod' accessing other underlay addresses, not changed.
2. Underlay pod to access overlay excluded ip addresses: this traffic should never pass through vxlan interface.
3. Overlay pod to access overlay excluded ip addresses: the same with overlay pod's accessing ip addresses outside the cluster, should never pass through vxlan interface.
4. Overlay pod to access underlay excluded ip addresses: the same with overlay pod's accessing ip addresses outside the cluster, should never pass through vxlan interface.

So, in the PR, I add some `throw` routes to handle the traffic with excluded destination address blocks, to make traffic don't pass through the vxlan interface if either it's destination is a overlay excluded ip addresses or it's the traffic from overlay pod to underlay excluded ip addresses.

### Describe how to verify it
Use overlay pod to access underlay gateway or excluded ip addresses.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/containernetwork/utils.go
+++ b/pkg/daemon/containernetwork/utils.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	daemonutils "github.com/oecp/rama/pkg/daemon/utils"
 
@@ -234,4 +235,70 @@ func ListLocalAddressExceptLink(exceptLinkName string) ([]netlink.Addr, error) {
 
 func CheckIPIsGlobalUnicast(ip net.IP) bool {
 	return !ip.IsInterfaceLocalMulticast() && ip.IsGlobalUnicast()
+}
+
+func checkPodRuleExist(podCidr *net.IPNet, family int) (bool, error) {
+	ruleList, err := netlink.RuleList(family)
+	if err != nil {
+		return false, fmt.Errorf("list rule failed: %v", err)
+	}
+
+	for _, rule := range ruleList {
+		if rule.Src != nil && podCidr.String() == rule.Src.String() {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func checkPodNeighExist(podIP net.IP, forwardNodeIfIndex int, family int) (bool, error) {
+	neighList, err := netlink.NeighProxyList(forwardNodeIfIndex, family)
+	if err != nil {
+		return false, fmt.Errorf("list neighs for forward node if index %v failed: %v", forwardNodeIfIndex, err)
+	}
+
+	for _, neigh := range neighList {
+		if neigh.IP.Equal(podIP) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func checkPodNetConfigReady(podIP net.IP, podCidr *net.IPNet, forwardNodeIfIndex int, family int) error {
+	backOffBase := 100 * time.Microsecond
+	retries := 4
+
+	for i := 0; i < retries; i++ {
+		time.Sleep(backOffBase)
+		backOffBase = backOffBase * 2
+
+		neighExist, err := checkPodNeighExist(podIP, forwardNodeIfIndex, family)
+		if err != nil {
+			return fmt.Errorf("check pod ip %v neigh exist failed: %v", podIP, err)
+		}
+
+		ruleExist, err := checkPodRuleExist(podCidr, family)
+		if err != nil {
+			return fmt.Errorf("check cidr %v rule exist failed: %v", podCidr, err)
+		}
+
+		if neighExist && ruleExist {
+			break
+		}
+
+		if i == retries-1 {
+			if !neighExist {
+				return fmt.Errorf("proxy neigh for %v is not created, waiting for daemon to create it", podIP)
+			}
+
+			if !ruleExist {
+				return fmt.Errorf("policy rule for %v is not created, waiting for daemon to create it", podCidr)
+			}
+		}
+	}
+
+	return nil
 }
--- a/pkg/daemon/route/route.go
+++ b/pkg/daemon/route/route.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"net"
 
+	daemonutils "github.com/oecp/rama/pkg/daemon/utils"
+
 	"github.com/oecp/rama/pkg/daemon/iptables"
 
 	"github.com/vishvananda/netlink"
@@ -49,16 +51,6 @@ import (
 // Local pod direct table doesn't need to be maintained manually,
 // because route will be deleted by kernel while specific device is not exist.
 
-type SubnetInfo struct {
-	cidr              *net.IPNet
-	gateway           net.IP
-	forwardNodeIfName string
-	autoNatOutgoing   bool
-	isOverlay         bool
-}
-
-type SubnetInfoMap map[string]*SubnetInfo
-
 type Manager struct {
 	// Use fixed table num to mark "local pod direct rule"
 	localDirectTableNum int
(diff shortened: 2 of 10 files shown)


pkg/daemon/route/route.go
```diff
@@ -165,33 +161,67 @@ func CreateRouteManager(localDirectTableNum, toOverlaySubnetTableNum, overlayMar
...
+		subnetInfo.reservedIPs = append(subnetInfo.reservedIPs, reservedIPs...)
+	}
+
+	if len(excludeIPs) != 0 {
+		subnetInfo.excludeIPs = append(subnetInfo.excludeIPs, excludeIPs...)
+	}
+
+	if start != nil || end != nil {
+		if start == nil {
+			start = cidr.IP
```
mars1024, line 204: actually we are using the second IP of network IP as start

pkg/daemon/route/route.go
```diff
@@ -165,33 +161,67 @@ func CreateRouteManager(localDirectTableNum, toOverlaySubnetTableNum, overlayMar
...
+		}
+
+		if end == nil {
+			end = daemonutils.LastIP(cidr)
```
mars1024, line 208: you could use the functions as IPAM, no similar functions in same project

pkg/daemon/route/route.go
hhyasdf: This is used for calculate a CIDR, not available ip to allocate, every possible ip should be considered.

pkg/daemon/route/route.go
hhyasdf: For now, this part of functions is only used for parse an ip range in daemon, maybe is not necessary until we actually find it's a common usage.

pkg/daemon/utils/ip_range.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func calculateIPLastZeroBits(ip net.IP) int {
+	zeroBits := 0
+	for {
+		nextPossibleZeroBits := zeroBits + 1
+		ipInt := ipToInt(ip)
+		ipUint64 := ipInt.Uint64()
+
+		if ipUint64 == (ipUint64>>nextPossibleZeroBits)<<nextPossibleZeroBits {
```
mars1024, line 163: 这里没有必要每次搞个新的 ipUint64 啊，位信息还在，一直左移一位+比较就可以了

pkg/daemon/utils/ip_range.go
```diff
@@ -0,0 +1,226 @@
...
+		start: start,
+		end:   end,
+	}, nil
+}
+
+func (ir *IPRange) TryAddIP(ipAddr net.IP) (success bool) {
+	if ipAddr.Equal(ip.PrevIP(ir.start)) {
+		ir.start = ipAddr
+		return true
+	} else if ipAddr.Equal(ip.NextIP(ir.end)) {
```
mars1024, line 36: no need else here

pkg/daemon/utils/ip_range.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+
+		if ipRange != nil {
+			excludeIPRanges = append(excludeIPRanges, ipRange)
+		}
+	}
+
+	allExcludedIPs := excludeIPs
+	if len(reservedIPs) != 0 {
+		allExcludedIPs = append(allExcludedIPs, reservedIPs...)
```
mars1024, line 108: reservedIPs should be kept not excluded

pkg/daemon/utils/ip_range.go
```diff
@@ -0,0 +1,226 @@
...
+	if start.To4() == nil {
+		ipLen = net.IPv6len * 8
+	}
+
+	minCidrPrefixLen := ipLen - zeroBits
+	maxValidCidrPrefixLen := minCidrPrefixLen
+
+	for {
+		tmpCidr := &net.IPNet{
+			IP:   start,
```
mars1024, line 186: should IP be masked by Mask to be a standard IPNet?

pkg/daemon/route/utils.go
```diff
@@ -295,32 +302,106 @@ func ensureFromPodSubnetRuleAndRoutes(forwardNodeIfName string, cidr *net.IPNet,
...
+		localAddrList, err := netlink.AddrList(nil, family)
+		if err != nil {
+			return fmt.Errorf("list local addresses failed: %v", err)
+		}
+
+		isLocalSubnet := false
+		for _, address := range localAddrList {
+			if cidr.Contains(address.IP) {
+				// Check if address is an enhanced address or used to connect a subnet.
+				if address.Flags&unix.IFA_F_NOPREFIXROUTE == 0 {
```
mars1024, line 322: if this judgment locate the local address except enhanced addresses? why?

pkg/daemon/utils/ip_range.go
hhyasdf: ReservedIPs can not be allocated to pod either, and also can be used to "exclude" ip, so I think is better to keep it.

pkg/daemon/route/utils.go
hhyasdf: Enhanced addresses will always have the `IFA_F_NOPREFIXROUTE` flag.

pkg/daemon/utils/ip_range.go
hhyasdf: No need to do this, the last `zeroBits` must be zero, we don't need to mask it again.

pkg/daemon/utils/ip_range.go
hhyasdf: fixed

pkg/daemon/utils/ip_range.go
hhyasdf: fixed

pkg/daemon/utils/ip_range.go
mars1024: Mask is changing every loop, so we could get some nonstandard CIDR like 192.168.0.4/24, 192.168.0.4/25, 192.168.0.4/26 ...

pkg/daemon/utils/ip_range.go
```diff
@@ -0,0 +1,227 @@
+package utils
```
mars1024, line 1: missing copyright here

pkg/daemon/containernetwork/utils.go
```diff
@@ -235,3 +235,55 @@ func ListLocalAddressExceptLink(exceptLinkName string) ([]netlink.Addr, error) {
...
+	for _, neigh := range neighList {
+		if neigh.IP.Equal(podIP) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func checkPodNetConfigReady(podIP net.IP, podCidr *net.IPNet, forwardNodeIfIndex int, family int) error {
```
mars1024, line 269: should we retry some times here to wait an async ready? I don't think it is good to let kubelet throw out the unnecessary errors.

pkg/daemon/utils/ip_range.go
mars1024: mistake, gotcha your point

pkg/daemon/containernetwork/utils.go
hhyasdf: fixed

pkg/daemon/utils/ip_range.go
hhyasdf: fixed

mars1024 (approved):
/lgtm
/approve
Thanks!

mars1024:
ref to #30

PR #39, opened 2021-08-03
enable daemon for inter-cluster pod communication config

SwimilTylers:
Pull Request Description
---

enable rama daemon to configure routes for pod-to-pod coms across clusters

NONE

### Describe how you did it

+ add routes for remote subnets
+ add userspace arp proxy for remote nodes and remote pods
+ add remote vtep to vxlan fdb
+ allow rama firewall to pass through inter-cluster overlay flows

### Special notes for reviews


pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,149 @@ type NetworkList struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   RemoteClusterSpec   `json:"spec"`
+	Status RemoteClusterStatus `json:"status"`
+}
+
+// RemoteClusterSpec is the spec for a RemoteCluster resource
+type RemoteClusterSpec struct {
+	// globally unique id
+	ClusterID uint32 `json:"cluster_id"`
```
hhyasdf, line 196: why not just use cluster name for globally unique id?

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,149 @@ type NetworkList struct {
...
+	// just for human memory, no practical use
+	ClusterName string              `json:"cluster_name"`
```
hhyasdf, line 198: metav1.TypeMeta already has a 'Name' filed

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,149 @@ type NetworkList struct {
...
+	Spec   RemoteSubnetSpec   `json:"spec"`
+	Status RemoteSubnetStatus `json:"status"`
+}
+
+// RemoteSubnetSpec is the spec for a RemoteSubnet resource
+type RemoteSubnetSpec struct {
+	Version      IPVersion   `json:"version"`
+	CIDR         string      `json:"cidr"`
+	Type         NetworkType `json:"type,omitempty"`
+	ClusterID    uint32      `json:"cluster_id"`
```
hhyasdf, line 268: ditto

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,149 @@ type NetworkList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   RemoteVtepSpec   `json:"spec"`
+	Status RemoteVtepStatus `json:"status"`
+}
+
+// RemoteVtepSpec is the spec for a RemoteVtep resource
+type RemoteVtepSpec struct {
+	ClusterID uint32 `json:"cluster_id"`
```
hhyasdf, line 303: ditto

pkg/apis/networking/v1/types.go
Louhwz: I prefer using cluster name as a alias, and it can be easily modified. ClusterID is somewhat like a constant, and shouldn't   make change during runtime.

pkg/apis/networking/v1/types.go
Louhwz: yes, clustername is redundant. I prefer deleting it.

pkg/daemon/controller/remote.go
```diff
@@ -0,0 +1,177 @@
...
+func (c *Controller) enqueueAddOrDeleteRemoteVtep(obj interface{}) {
+	c.nodeQueue.Add(ActionReconcileNode)
+}
+
+func (c *Controller) enqueueUpdateRemoteVtep(oldObj, newObj interface{}) {
+	oldRv := oldObj.(*ramav1.RemoteVtep)
+	newRv := newObj.(*ramav1.RemoteVtep)
+
+	if oldRv.Spec.VtepIP != newRv.Spec.VtepIP ||
+		oldRv.Spec.VtepMAC != newRv.Spec.VtepMAC || !reflect.DeepEqual(oldRv.Status.PodIPList, newRv.Status.PodIPList) {
```
Louhwz, line 73: Slice is unsorted. If using deepequal, two slice having same value maybe false. Using set instead. For example, "github.com/gogf/gf/container/gset"

pkg/daemon/controller/remote.go
```diff
@@ -0,0 +1,177 @@
...
+type RemoteVtepCache struct {
+	remoteVtepIPMap map[string]net.HardwareAddr
+
+	ch chan struct{}
+}
+
+func NewRemoteVtepCache() *RemoteVtepCache {
+	return &RemoteVtepCache{
+		remoteVtepIPMap: map[string]net.HardwareAddr{},
+		ch:              make(chan struct{}, 1),
```
Louhwz, line 24: buffered channel is Non-blocking, this lock can't play a role. Use sync.RWMutex{} instead.

SwimilTylers: I just follow the code style of NodeCache, but I think sync.RWMutex is a good idea.

pkg/daemon/controller/remote.go
SwimilTylers: Nice I've thought about that before. Is reordering of pod ip list a frequent op? If not, I believe deepEqual is tolerable.

pkg/daemon/controller/remote.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func (rvc *RemoteVtepCache) UpdateRemoteVtepIPs(remoteVteps []*ramav1.RemoteVtep) error {
+	rvc.lock()
+	defer rvc.unlock()
```
Louhwz, line 30: Don't lock in such a big range, this will slow the speed. Use lock in a smaller range. Like extraing a function called "setRemoteVtep", and lock in that function.

pkg/apis/networking/v1/types.go
hhyasdf: > I prefer using cluster name as a alias, and it can be easily modified. ClusterID is somewhat like a constant, and shouldn't make change during runtime.

If use 'Name' filed in metav1.TypeMeta as cluster name, it's also unchangeable. And same thing happens to other cluster-scoped k8s object too, e.g., name of a 'Node' cannot be edited.

pkg/daemon/controller/remote.go
Louhwz: Reordering cannot be guaranteed. Besides, reflect is slow. Try the set and use diff fucntion, it's efficient.

SwimilTylers: OK, I shall follow that.

pkg/daemon/controller/remote.go
Louhwz: which nodecache?

pkg/daemon/controller/remote.go
Louhwz: why? i can't understand, pls explain more

pkg/apis/networking/v1/types.go
Louhwz: I see. I will delete both clusterID and clusterName.

pkg/daemon/controller/remote.go
SwimilTylers: `NodeIPCache` in `pkg/daemon/controller/node.go`. Anyway, I shall take your advice.

pkg/daemon/controller/remote.go
SwimilTylers: Increasing the capacity of `map` involves rehasing.

pkg/daemon/controller/remote.go
Louhwz: ok

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,149 @@ type NetworkList struct {
 
 	Items []Network `json:"items"`
 }
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
mars1024, line 181: missing subresource

PR #40, opened 2021-08-04
improvement: clear-stale-neigh-entries-for-overlay

hhyasdf:
Pull Request Description
---

Consume there is a node A (it's the same thing for pod on this node) has accessed pod B, there will be a REACHABLE neigh entry for pod B on node A. If the pod B is deleted, and after a period of time, neigh entry for pod B on node A will become STALE, even pod B has been recreate on another different node, this neigh entry will not be updated, because there is no gratuitous arp/ndp for overlay network.

This is because kernel will never clear a STALE neigh entry if the number of all neigh entries on this node is under the limit of gc_thresh parameter, which will cause a problem of seconds packet loss while this node want to rebuild a connection to the new pod B.

NONE

### Describe how you did it
Clear STALE neigh entries for vxlan interface once the rama-daemon pod is running, and clear specific STALE neigh if the ip is not found by neigh proxy.

### Describe how to verify it
1. recreate rama-daemon pod to clear STALE neigh entries.
2. delete an exist pod and check if the STALE neigh cache will be cleaned for all nodes.

### Special notes for reviews

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,3 +29,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed Issues
 - Fix error data path for overlay pod to access underlay gateway and excluded ip addresses
+
+## v0.1.2
+### Improvements
+- Clear stale neigh entries for overlay network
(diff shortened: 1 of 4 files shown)


mars1024 (approved):
/lgtm, good job!
/approve

PR #42, opened 2021-08-05
multicluster crd

Louhwz:
Pull Request Description
---

1. Add crd about remot cluster
2. Add code-gen target in makefile
3. code-gen


pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   RemoteSubnetSpec   `json:"spec"`
+	Status RemoteSubnetStatus `json:"status"`
+}
+
+// RemoteSubnetSpec is the spec for a RemoteSubnet resource
+type RemoteSubnetSpec struct {
+	Version      IPVersion   `json:"version"`
+	CIDR         string      `json:"cidr"`
```
hhyasdf, line 261: Cannot just use a CIDR string to describe a rama ip address range, rama supports a discontinuous ip address range description. (excludeIPs, start, end)

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	Type         NetworkType `json:"type"`
+	ClusterName  string      `json:"clusterName"`
+	OverlayNetID *uint32     `json:"overlayNetID,omitempty"`
```
hhyasdf, line 264: If underlay network will also have a RemoteSubnet CR, maybe 'NetID' is better than 'OverlayNetID'.

pkg/apis/networking/v1/types.go
Louhwz: In my opinion, ip address range description is not needed. Remote subnet and subnet are one-to-many relationship，and we only need the cidr for 40000 and 10000+ route table, right? @SwimilTylers

Louhwz: excludeIPs won't be recorded in "podiplist", reserved ip too. @hhyasdf

SwimilTylers: I guess it is caused by latest pr of `main`. We shall discuss about it @Louhwz .

hhyasdf: > In my opinion, ip address range description is not needed. Remote subnet and subnet are one-to-many relationship，and we only need the cidr for 40000 and 10000+ route table, right? @SwimilTylers

For local overlay subnet to access remote underlay subnet, if an ip address is excluded in remote underlay subnet, this ip address should be handled as an external ip, which is not in both local and remote cluster. So we need the information of excluded ip (and 'start', 'end').

hhyasdf: > excludeIPs won't be recorded in "podiplist", reserved ip too. @hhyasdf

Information of "podiplist" is used to assist the process of kernel vxlan neigh resolving, without it, kernel vxlan neigh resolving will also work out by policy routes. So accurate policy routes is needed.

Louhwz: I got the meaning of "excluded ip", but why do we need 'start' and 'end'?

hhyasdf: > I got the meaning of "excluded ip", but why do we need 'start' and 'end'? @hhyasdf

We need them all. IP addresses beyond the range of 'start' and 'end' are also excluded ip addresses.

hhyasdf: > I got the meaning of "excluded ip", but why do we need 'start' and 'end'? @hhyasdf

'excluedIPs' is not the same with 'excluded ip addresses'.

pkg/apis/networking/v1/types.go
Louhwz: First, this is pointer, if it is underlay network, it is nil and will not be printed.
Second, this is actually a vxlan network identifier in all connected domain, I this OverlayNetID is more clear @hhyasdf

pkg/apis/networking/v1/types.go
Louhwz: > > I got the meaning of "excluded ip", but why do we need 'start' and 'end'? @hhyasdf
> 
> 'excludeIPs' is not the same with 'excluded ip addresses'.

yeah, i got this.

Louhwz: ok, i see. 3q

pkg/apis/networking/v1/types.go
hhyasdf: If it is actually a vxlan network identifier in all connected domain, I think maybe it's a cluster-scoped information, why not keep it in RemoteCluster CR?

Louhwz: Sound good to me

Louhwz:
Updated, pls check @hhyasdf. Just cut in, after the conversation with Yi-Jiang Yang , I used addressrange directly in "remotesubnet"

hhyasdf (commented):
/lgtm, cc @mars1024

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
+
+// RemoteCluster is a specification for a RemoteCluster resource
+type RemoteCluster struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   RemoteClusterSpec   `json:"spec"`
+	Status RemoteClusterStatus `json:"status"`
```
mars1024, line 190: every unnecessary field should be tagged with omitempty, otherwise encoding or decoding will meet some unexpected errors

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+}
+
+// RemoteClusterSpec is the spec for a RemoteCluster resource
+type RemoteClusterSpec struct {
+	ConnConfig APIServerConnConfig `json:"connConfig"`
```
mars1024, line 195: every field need a comment

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	ClientKey  []byte `json:"clientKey"`
+	// The maximum length of time to wait before giving up on a server
+	// request. A value of zero means no timeout. Default: zero second
+	Timeout uint32 `json:"timeout"`
+}
+
+// RemoteClusterStatus is the status for a RemoteCluster resource
+type RemoteClusterStatus struct {
+	// Conditions is an array of current cluster conditions.
+	Conditions []ClusterCondition `json:"conditions"`
```
mars1024, line 212: why this is slice?

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	// If network in remote cluster is all underlay, it is nil
+	TunnelNetID *uint32 `json:"tunnelNetID,omitempty"`
```
mars1024, line 214: why this field in status? how to fetch it?

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+
+	Items []RemoteCluster `json:"items"`
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
```
mars1024, line 249: ditto

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+type RemoteVtepSpec struct {
+	ClusterName string `json:"clusterName"`
+	NodeName    string `json:"nodeName"`
+	VtepIP      string `json:"vtepIP"`
+	VtepMAC     string `json:"vtepMAC"`
+}
+
+// RemoteVtepStatus is the status for a RemoteVtep resource
+type RemoteVtepStatus struct {
+	PodIPList      []string    `json:"podIPList"`
```
mars1024, line 306: this should be in spec not status

pkg/apis/networking/v1/types.go
Louhwz: List from Remote cluster's overlay network. Every Health check will do the job.

pkg/apis/networking/v1/types.go
Louhwz: It is not applied to other crd like ipinstance, subnet, and i think "ConnConfig" is enough to explain the usage.  
Code as comment.

pkg/apis/networking/v1/types.go
Louhwz: Copy from kube-fed. When a cluser is not ready, there are many reasons. For example, if the /healthz returns nil error, however, it's body is not "ok", we should set the Status like [Ready=false, Reachable=true], what's more, we can also check if the remote cluster has defined rama cr.

pkg/apis/networking/v1/types.go
Louhwz: I just follow the other definition, they don't set omitempty in Status. What's more, omitempty for An embedded struct  is useless.

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
 
 	Items []Network `json:"items"`
 }
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:openapi-gen=true
```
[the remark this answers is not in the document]
Louhwz: I just follow other definition. Why do we need subresource?

pkg/apis/networking/v1/types.go
Louhwz: @mars1024

pkg/apis/networking/v1/types.go
mars1024: change this to slice

pkg/apis/networking/v1/types.go
```diff
@@ -174,3 +175,144 @@ type NetworkList struct {
...
+	Status RemoteVtepStatus `json:"status"`
+}
+
+// RemoteVtepSpec is the spec for a RemoteVtep resource
+type RemoteVtepSpec struct {
+	ClusterName string   `json:"clusterName"`
+	NodeName    string   `json:"nodeName"`
+	VtepIP      string   `json:"vtepIP"`
+	VtepMAC     string   `json:"vtepMAC"`
+	PodIPList   []string `json:"podIPList"`
```
mars1024, line 302: rename PodIPList to IPList

SwimilTylers: Is IPList rather ambiguous?

hhyasdf: Maybe EndpointIPList is okay, ip addresses for vxlan endpoints.

mars1024 (approved):
/approve

PR #43, opened 2021-08-06
add cluster uuid &&  rename ip list to endpoint ip list

Louhwz:
Pull Request Description
---

1. Add uuid to identify each cluster. UUID is useful when identifying remote cluster, and because there are no centralized organizer to dispatch metadata, we need to set it by directly using the corresponding remote cluster's namespace "kube-system" uid.
2. The field IpList in remotevtep is too ambiguous, change it to endpoint ip list will make it more clear
I also add a comment to explain TunnelNetID
3. Delete tunnel net  id
### Special notes for reviews


pkg/apis/networking/v1/types.go
```diff
@@ -260,7 +260,8 @@ type RemoteSubnetSpec struct {
 	Range       AddressRange `json:"range"`
 	Type        NetworkType  `json:"type"`
 	ClusterName string       `json:"clusterName"`
-	TunnelNetID *uint32      `json:"tunnelNetID"`
+	// If the corresponding subnet is in underlay network, it will be set to nil
```
hhyasdf, line 263: It should not be nil for a remote underlay subnet either, because a pod from local overlay subnet need to access remote underlay subnet through a tunnel.

hhyasdf: > If it is nil, use local vxlan net id.

For this case, local vxlan net id is the tunnelNetID, maybe tunnelNetID should not be nil.

hhyasdf: Do not hide information from user if there is already a filed to show it. Just show it as directly as possible.

Louhwz: > > If it is nil, use local vxlan net id.
> 
> For this case, local vxlan net id is the vxlan id used to tunneling, maybe tunnelNetID should not be nil either.

you mean the control plane should perceive the local vxlan net id? maybe too coupled?

hhyasdf: > > > If it is nil, use local vxlan net id.
> > 
> > 
> > For this case, local vxlan net id is the vxlan id used to tunneling, maybe tunnelNetID should not be nil either.
> 
> you mean the control plane should perceive the local vxlan net id? maybe too coupled?

You don't need to make it separated, whether use local vxlan net id or not, it's an intact logic. You always need to looking for the remote overlay network, the only thing you need to do is to look back to local overlay network if the remote overlay network not exist.

From another side, tunnelNetID is the actual vxlan id for tunneling, this is a good information for network operation engineer, we should not just make it nil if we have an vxlan id for it, whether the vlxan id is local vxlan id or not.

Louhwz:
pls check, @hhyasdf

hhyasdf (approved):
/lgtm, thanks

hhyasdf (approved):

Louhwz:
@hhyasdf pls check

pkg/apis/networking/v1/types.go
```diff
@@ -210,6 +211,10 @@ type APIServerConnConfig struct {
 type RemoteClusterStatus struct {
 	// Conditions is an array of current cluster conditions.
 	Conditions []ClusterCondition `json:"conditions"`
+	// A globally unique identifier. Use for validating.
+	// Generated by directly using the corresponding remote cluster's namespace "kube-system" uid
+	// types.UID is also universally unique identifiers (also known as UUIDs).
+	UUID types.UID `json:"uuid"`
```
mars1024, line 217: I think `string` is more general here?

Louhwz: types.UID is actually a type definition of string, which means it is convenient to convert from string or convert to string, what's more, it is used in kubernetes, i think directly use it is more understandable.

mars1024 (approved):
/lgtm

PR #44, opened 2021-08-09
control plane about remote cluster

Louhwz:
Pull Request Description
---

1. It implements the control plane about remote cluster.
### Describe how you did it
1. Add remote cluster controller
2. To do cluster management, like health checking, cluster discovery, I also add a remote cluster manager.
3. Model converting between existing rama cr and added remote cr.
### Special notes for reviews


Dockerfile.amd64
```diff
@@ -1,11 +1,17 @@
 FROM golang:1.13.8 as builder
 
-COPY . /go/src/github.com/oecp/rama
-
-ENV ARCH=amd64
+ENV ARCH=arm64
```
hhyasdf, line 3: "ARCH" needs to be "amd64" ?

Louhwz: sorry for that, committed.

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,206 @@
...
+	for m.processNextNode() {
+	}
+}
+
+func (m *Manager) filterNode(obj interface{}) bool {
+	if !m.GetIsReady() {
+		return false
+	}
+	// todo debug
+	klog.Infof("[RemoteCluster-Node]debug. node=%v", utils.ToJSONString(obj))
```
hhyasdf, line 178: If it's only for develop usage, maybe need to be removed.

pkg/webhook/validating/network.go
```diff
@@ -124,5 +124,16 @@ func NetworkDeleteValidation(ctx context.Context, req *admission.Request, handle
 			return admission.Denied(fmt.Sprintf("still have child subnet %s", subnet.Name))
 		}
 	}
+
+	if network.Spec.Type == ramav1.NetworkTypeOverlay && feature.MultiClusterEnabled() {
+		remoteClusterLister := &ramav1.RemoteClusterList{}
```
hhyasdf, line 129: Lister has a special meaning in k8s, `remoteClusterLister` should rename to `remoteClusterList`

pkg/webhook/validating/remotecluster.go
```diff
@@ -0,0 +1,85 @@
...
+
+	ramav1 "github.com/oecp/rama/pkg/apis/networking/v1"
+	"github.com/oecp/rama/pkg/utils"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var (
+	r                = regexp.MustCompile(`^(https?://)[\w-]+(\.[\w-]+)+:\d{1,5}$`)
```
hhyasdf, line 18: "r" is not a proper variable name.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,209 @@
...
+	ans := make([]*networkingv1.IPInstance, 0)
+	for _, v := range podIP {
+		if instance, ok := v.(*networkingv1.IPInstance); ok {
+			ans = append(ans, instance)
+		}
+	}
+	return ans, nil
+}
+
+func (m *Manager) nodeToIPList(nodeName string) ([]string, error) {
```
hhyasdf, line 118: Without clustermesh or not, daemon will update node local ip addresses used for rama overlay network to `node.Annotations[constants.AnnotationNodeLocalVxlanIPList]`, which should also be updated to spec.endpointIPList of RemoteVtep CRs.

pkg/controller/ipam/network.go
```diff
@@ -35,23 +34,13 @@ func (c *Controller) filterNetwork(obj interface{}) bool {
 }
 
 func (c *Controller) addNetwork(obj interface{}) {
-	n, ok := obj.(*v1.Network)
-	if !ok {
```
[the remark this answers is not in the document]
Louhwz: check k8s.io/client-go/tools/cache/controller.go:246

pkg/controller/remotecluster/cache.go
```diff
@@ -0,0 +1,37 @@
+package remotecluster
+
+import (
+	"sync"
+
+	"github.com/oecp/rama/pkg/rcmanager"
+	"k8s.io/klog"
+)
+
+type Cache struct {
```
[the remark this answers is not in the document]
Louhwz: You are right

pkg/controller/remotecluster/cache.go
```diff
@@ -0,0 +1,37 @@
...
+
+	c.rcMgrMap[clusterName] = manager
+}
+
+func (c *Cache) Del(clusterName string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if rc, exists := c.rcMgrMap[clusterName]; exists {
+		klog.Infof("Delete cluster %v from cache", clusterName)
+		close(rc.StopCh)
```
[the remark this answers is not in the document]
Louhwz: ok, thanks

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,127 @@
...
+func (c *Controller) startRemoteClusterMgr(clusterName string) error {
+	klog.Infof("[debug] processNextRemoteClusterMgr name=%v", clusterName)
+	rcManager, exists := c.rcMgrCache.Get(clusterName)
+	if !exists {
+		klog.Errorf("Can't find rcManager. clusterName=%v", clusterName)
+		return errors.Errorf("Can't find rcManager. clusterName=%v", clusterName)
+	}
+	klog.Infof("Start single remote cluster manager. clusterName=%v", clusterName)
+
+	managerCh := rcManager.StopCh
```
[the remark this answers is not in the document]
Louhwz: Absolutely right.

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,127 @@
...
+
+// use remove+add instead of update
+func (c *Controller) addOrUpdateRCMgr(rc *networkingv1.RemoteCluster) error {
+	// lock in function range to avoid renewing cluster manager when newing one
+	c.rcMgrCache.mu.Lock()
+	defer c.rcMgrCache.mu.Unlock()
+	klog.Infof("[addOrUpdateRCMgr] cluster=%v", rc.Name)
+
+	clusterName := rc.Name
+	if k, exists := c.rcMgrCache.rcMgrMap[clusterName]; exists {
```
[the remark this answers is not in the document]
Louhwz: I think there is no need to do it. First, remote cluster manager will only be updated if users change remote cluster's cr which has very low frequency. Second, if user change remote cluster connection config, it implies old connection is outdated, we need to create a new connection.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,209 @@
...
+	}(obj)
+
+	if err != nil {
+		klog.Error(err)
+	}
+	return true
+}
+
+func (m *Manager) filterIPInstance(obj interface{}) bool {
+	if !m.GetIsReady() {
```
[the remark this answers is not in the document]
Louhwz: I wrote this deliberately. In my opinion, if remote cluster is not ready, retry is meaningless, we can drop the event directly. After the remote cluster is ready in health checking, it will trigger a reconcile about remote subnet and remote vtep.

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return err
+	}
+
+	add, update, remove := m.diffNodeAndVtep(nodes, vteps)
+	var (
+		wg  sync.WaitGroup
+		cur = metav1.Now()
+	)
+	wg.Add(3)
```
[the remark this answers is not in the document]
Louhwz: https://stackoverflow.com/questions/65213707/where-to-put-wg-add

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,206 @@
...
+
+}
+
+func (m *Manager) RunNodeWorker() {
+	for m.processNextNode() {
+	}
+}
+
+func (m *Manager) filterNode(obj interface{}) bool {
+	if !m.GetIsReady() {
```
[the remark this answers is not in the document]
Louhwz: ditto

pkg/rcmanager/subnet.go
```diff
@@ -0,0 +1,300 @@
...
+			networkMap[network.Name] = network
+		}
+		return networkMap
+	}()
+	add, update, remove := m.diffSubnetAndRCSubnet(remoteClusterSubnets, localClusterRemoteSubnets, networkMap)
+	var (
+		wg  sync.WaitGroup
+		cur = metav1.Now()
+	)
+	wg.Add(3)
```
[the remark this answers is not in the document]
Louhwz: https://stackoverflow.com/questions/65213707/where-to-put-wg-add

pkg/rcmanager/subnet.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	localClusterRemoteSubnets, err := m.remoteSubnetLister.List(labels.NewSelector())
+	if err != nil {
+		return err
+	}
+	remoteClusterSubnets, err := m.subnetLister.List(labels.NewSelector())
+	if err != nil {
+		return err
+	}
+	networks, err := m.networkLister.List(labels.NewSelector())
```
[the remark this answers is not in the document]
Louhwz: 555,i know that

yamls/rama-init.yaml
```diff
@@ -214,4 +220,91 @@ spec:
...
+spec:
+  group: networking.alibaba.com
+  version: v1
+  scope: Cluster
+  names:
+    plural: remoteclusters
+    singular: remotecluster
+    kind: RemoteCluster
+    shortNames:
+      - rcc
```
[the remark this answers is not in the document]
Louhwz: Rmotecluster is so long...If there are no abbreviation, i think is not convecient to use. Also, "rcc" stands for "Remote Cluster's Cluster", @mars1024 check it.

cmd/webhook/main.go
```diff
@@ -40,6 +40,7 @@ var (
 	scheme             = runtime.NewScheme()
 	port               int
 	metricsBindAddress string
+	stopCh             = make(chan struct{})
```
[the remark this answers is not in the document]
Louhwz: how about controllerruntime.SetupSignalHandler()? `InformerFactory` needs a stop chan

pkg/controller/remotecluster/cache.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"sync"
+
+	"github.com/oecp/rama/pkg/rcmanager"
+	"k8s.io/klog"
+)
+
+type Cache struct {
+	mu       sync.RWMutex
```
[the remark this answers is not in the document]
Louhwz: I don't think it is too different. You can say pointer is faster when copying, but it is allocated on heap and thus have GC pressure. Anyway, I see your code in pkg/controller/ipam/cache.go, I will follow it.

pkg/rcmanager/subnet.go
```diff
@@ -0,0 +1,300 @@
...
+			subnetMap[s.Name] = s
+		}
+		return subnetMap
+	}()
+
+	for _, v := range rcSubnets {
+		if v.ClusterName != m.ClusterName {
+			continue
+		}
+		if subnet, exists := subnetMap[v.Name]; !exists {
```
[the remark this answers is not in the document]
Louhwz: Yes, Remote Cluster's name can't be empty string and remote subnet's name format is "%v.%v"

pkg/utils/remotesubnet.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+import (
+	"fmt"
+)
+
+const remoteSubnetNameFormat = "%v.%v"
```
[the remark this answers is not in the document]
Louhwz: I think it is not bad because subnet is already separated by "-", using another separator can make the name more precise. For example, `"remotecluster-2.overlay-subnet1"` vs `"remotecluster-2-overlay-subnet1"`

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,127 @@
...
+
+	_, err = c.ramaClient.NetworkingV1().RemoteClusters().UpdateStatus(context.TODO(), rc, metav1.UpdateOptions{})
+	if err != nil {
+		runtime.HandleError(err)
+		return err
+	}
+	rcMgr.SetIsReady(IsReady(conditions))
+
+	c.rcMgrCache.rcMgrMap[clusterName] = rcMgr
+	c.rcMgrQueue.Add(clusterName)
```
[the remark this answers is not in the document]
Louhwz: Because I want to decouple the creation and start up of remote cluster manager for the start up may be heavy

cmd/webhook/main.go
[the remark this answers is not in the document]
hhyasdf: > How about controllerruntime.SetupSignalHandler()? `InformerFactory` needs a stop chan

Yeah, maybe `controllerruntime.SetupSignalHandler()` is a better way to do this, rather than doing some extra channel-close logics.

Dockerfile.amd64
```diff
@@ -1,11 +1,17 @@
...
 
-COPY . /go/src/github.com/oecp/rama
-
 ENV ARCH=amd64
 
 WORKDIR /go/src/github.com/oecp/rama
 
+COPY go.mod ./go.mod
+COPY go.sum ./go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
```
mars1024, line 9: what brings a re-download problem here? I even don't think there it is.

Dockerfile.arm64
```diff
@@ -1,11 +1,17 @@
...
 
-COPY . /go/src/github.com/oecp/rama
-
 ENV ARCH=arm64
 
 WORKDIR /go/src/github.com/oecp/rama
 
+COPY go.mod ./go.mod
+COPY go.sum ./go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
```
mars1024, line 9: ditto

cmd/webhook/main.go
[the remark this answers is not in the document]
mars1024: Why we need a informer in webhook? I prefer what you want is a cacher not a controller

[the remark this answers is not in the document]
mars1024: BTW, controller-runtime provide us some useful utility packages about this, try them!

pkg/constants/labels.go
```diff
@@ -30,6 +31,8 @@ const (
 	LabelDualStackAddressQuota = "networking.alibaba.com/dualstack-address-quota"
 
 	LabelNetworkType = "networking.alibaba.com/network-type"
+
+	LabelRemoteVtepNodeName = "networking.alibaba.com/node-name"
```
mars1024, line 35: should we reuse `LabelNode` instead of creating a new one?

pkg/controller/ipam/network.go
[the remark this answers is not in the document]
mars1024: check https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L246 for what？

pkg/controller/remotecluster/cache.go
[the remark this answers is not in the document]
mars1024: use `sync.Map` here, a simple but useful utility package

Dockerfile.amd64
Louhwz: It happens when you are developing. At that time, re-download go module is  time-consuming. And this little trick can help speed the build of image.

pkg/controller/remotecluster/cache.go
```diff
@@ -0,0 +1,44 @@
...
+
+	c.rcMgrMap[clusterName] = manager
+}
+
+func (c *Cache) Del(clusterName string) {
+	c.RWMutex.Lock()
+	defer c.RWMutex.Unlock()
+	if rc, exists := c.rcMgrMap[clusterName]; exists {
+		klog.Infof("Delete cluster %v from cache", clusterName)
+		rc.Close()
```
mars1024, line 41: why deleting a cluster from cache will close it firstly? this sequential actions should be guaranteed by caller

pkg/controller/ipam/network.go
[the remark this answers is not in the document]
Louhwz: @mars1024 FilterFunction will first be called, thus it's redundant to check type conversion again.

pkg/controller/remotecluster/cache.go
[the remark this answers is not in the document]
Louhwz: sync.Map can only lock in `get` and `set` function, and sometimes we need to lock in a bigger range. Check the `addOrUpdateRCMgr`

pkg/controller/remotecluster/cache.go
Louhwz: What do  you mean, bro? Is it bad close in this sequence?

pkg/controller/remotecluster/rcmgr.go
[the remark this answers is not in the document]
hhyasdf: yep, copy that.

cmd/webhook/main.go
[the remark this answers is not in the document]
Louhwz: > Why we need a informer in webhook? I prefer what you want is a cacher not a controller

Yes, I need a cache. Are there any better solution?

pkg/controller/ipam/network.go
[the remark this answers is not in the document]
Louhwz: @mars1024 
https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L260
https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L268
https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L284

pkg/utils/ip.go
```diff
@@ -16,18 +16,33 @@
...
-// If IP is valid, return itself otherwise empty string
+// NormalizedIP If IP is valid, return itself otherwise empty string
 func NormalizedIP(ip string) string {
 	if net.ParseIP(ip) != nil {
 		return ip
 	}
 	return ""
 }
+
+func Intersect(cidr1 string, ipVersion1 networkingv1.IPVersion,
```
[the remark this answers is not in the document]
Louhwz: I gotcha

pkg/controller/remotecluster/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	<-stopCh
+
+	c.closeRemoteClusterManager()
+
+	klog.Info("Shutting down workers")
+	return nil
+}
+
+func (c *Controller) closeRemoteClusterManager() {
+	// no need to lock
```
mars1024, line 146: why no need to lock here? stopCh is just a signal, it can not make sure every working threads have already exited successfully

pkg/controller/remotecluster/controller.go
```diff
@@ -0,0 +1,250 @@
...
+
+	networks, err := c.localClusterNetworkLister.List(labels.Everything())
+	if err != nil {
+		klog.Warningf("Can't list local cluster network. err=%v", err)
+		return
+	}
+	for _, network := range networks {
+		if network.Spec.Type == networkingv1.NetworkTypeOverlay {
+			n := network.DeepCopy()
+			c.OverlayNetID = n.Spec.NetID
```
mars1024, line 164: loop here is not effective enough than events

pkg/controller/remotecluster/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	)
+	for _, rc := range remoteClusters {
+		r := rc.DeepCopy()
+		manager, exists := c.rcMgrCache.Get(r.Name)
+		if !exists {
+			continue
+		}
+		cnt = cnt + 1
+		wg.Add(1)
+		go c.updateSingleRCStatus(manager, r, &wg)
```
mars1024, line 191: use inline function here to avoid passing `wg` through functional parameters

pkg/controller/remotecluster/controller.go
Louhwz: Do we really need to lock for close? This close is used to close manager's informer factory, I think close it is enough. If locked, there maybe some delay because of the lock competion. What's more, even the goroutine exited failed, the main thread will do the clean job.

pkg/controller/remotecluster/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	remoteClusterInformer informers.RemoteClusterInformer,
+	remoteSubnetInformer informers.RemoteSubnetInformer,
+	localClusterSubnetInformer informers.SubnetInformer,
+	remoteVtepInformer informers.RemoteVtepInformer,
+	localClusterNetworkInformer informers.NetworkInformer) *Controller {
+	runtimeutil.Must(networkingv1.AddToScheme(scheme.Scheme))
+
+	klog.V(4).Info("Creating event broadcaster")
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(klog.Infof)
```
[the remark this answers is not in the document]
Louhwz: Well, I just follow `ipam controller`

pkg/controller/remotecluster/controller.go
Louhwz: why? It is pointer

pkg/controller/remotecluster/controller.go
Louhwz: I don't think there are too much difference for it uses the lister. By the way, events needs more code and is not as clear as single function.

pkg/rcmanager/ipinstance.go
Louhwz: yeah, i gotcha

pkg/rcmanager/node.go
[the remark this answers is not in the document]
hhyasdf: Well, it's a mistake to use it in a go routine. 

But actually what my point is, `wg.Add(3)` is not friendly for later maintenance, because it's always hard for someone to count how many goroutine we have in the codes, which might bring problems if a goroutine is actually added but without increase the number of `wg.Add(xxx)`.

Maybe write `wg.Add(1)` for every separate goroutine (like three times for three gorouine) just before the `go func() {}` is better.

pkg/controller/ipam/network.go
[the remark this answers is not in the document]
hhyasdf: Got it.

pkg/controller/remotecluster/rcmgr.go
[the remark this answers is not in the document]
hhyasdf: But it seems quite light now for starting a remote cluster manager, what `rcManager.Run()` do is just start a couple of goroutines and return.

And another benefit what I can think of it that if starting up a remote cluster manager is failed, we can restart it automatically using a work queue. But this will happen only if `rcMgrCache.Get()` is failed, which seems impossible.

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,132 @@
...
+	rc.Status.Conditions = conditions
+
+	_, err = c.ramaClient.NetworkingV1().RemoteClusters().UpdateStatus(context.TODO(), rc, metav1.UpdateOptions{})
+	if err != nil {
+		runtime.HandleError(err)
+		return err
+	}
+	rcMgr.SetIsReady(IsReady(conditions))
+
+	c.rcMgrCache.rcMgrMap[clusterName] = rcMgr
```
hhyasdf, line 79: Maybe use `rcMgrCache.Set()` instead.

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,132 @@
...
+
+// use remove+add instead of update
+func (c *Controller) addOrUpdateRCMgr(rc *networkingv1.RemoteCluster) error {
+	// lock in function range to avoid renewing cluster manager when newing one
+	c.rcMgrCache.RWMutex.Lock()
+	defer c.rcMgrCache.RWMutex.Unlock()
+	klog.Infof("[addOrUpdateRCMgr] cluster=%v", rc.Name)
+
+	clusterName := rc.Name
+	if mgr, exists := c.rcMgrCache.rcMgrMap[clusterName]; exists {
```
hhyasdf, line 52: Maybe use `rcMgrCache.Del()` instead.

pkg/controller/remotecluster/rcmgr.go
```diff
@@ -0,0 +1,132 @@
...
+		return errors.Errorf("Can't find rcManager. clusterName=%v", clusterName)
+	}
+	rcManager.Run()
+	return nil
+}
+
+// use remove+add instead of update
+func (c *Controller) addOrUpdateRCMgr(rc *networkingv1.RemoteCluster) error {
+	// lock in function range to avoid renewing cluster manager when newing one
+	c.rcMgrCache.RWMutex.Lock()
```
hhyasdf, line 47: Maybe there is not needed anymore if using `rcMgrCache.Del()` and. `rcMgrCache.Add()`.

pkg/utils/remotesubnet.go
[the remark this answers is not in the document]
hhyasdf: Got it.

pkg/rcmanager/ipinstance.go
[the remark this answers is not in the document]
hhyasdf: But the problem is ipinstance reconciling will not be triggered again?

pkg/rcmanager/subnet.go
```diff
@@ -0,0 +1,316 @@
...
+	if !m.GetIsReady() {
+		return false
+	}
+	_, ok := obj.(*networkingv1.Subnet)
+	return ok
+}
+
+func (m *Manager) addOrDelSubnet(obj interface{}) {
+	subnet, _ := obj.(*networkingv1.Subnet)
+	m.EnqueueSubnet(subnet.ObjectMeta.Name)
```
hhyasdf, line 267: If `subnet.ObjectMeta.Name` is not used, maybe use `rcmanager.ReconcileSubnet` is better.

pkg/rcmanager/subnet.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+func (m *Manager) updateSubnet(oldObj, newObj interface{}) {
+	oldRC, _ := oldObj.(*networkingv1.Subnet)
+	newRC, _ := newObj.(*networkingv1.Subnet)
+
+	if oldRC.ResourceVersion == newRC.ResourceVersion {
+		return
+	}
+	m.EnqueueSubnet(newRC.ObjectMeta.Name)
```
hhyasdf, line 277: ditto.

pkg/controller/remotecluster/cache.go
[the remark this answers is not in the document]
hhyasdf: Up to @mars1024 .

pkg/rcmanager/subnet.go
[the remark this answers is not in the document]
hhyasdf: I mean "_, exists := subnetMap[v.Name]" here will always return false, because the keys of `subnetMap` are `networkingv1.Subnet.Name` rather than `utils.GenRemoteSubnetName(m.ClusterName, subnet.Name)`.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,210 @@
...
+	return err
+}
+
+func (m *Manager) pickEndpointListFromNode(node *v1.Node) ([]string, error) {
+	instances, err := m.IPLister.List(labels.SelectorFromSet(labels.Set{constants.LabelNode: node.Name}))
+	if err != nil {
+		return nil, err
+	}
+	desired := utils.PickUsingIPList(instances)
+	nodeLocalVxlanipStringList := strings.Split(node.Annotations[constants.AnnotationNodeLocalVxlanIPList], ",")
```
hhyasdf, line 130: Considering it's a separate logic for daemon to config iptables rules for "node local vxlan ip", simply adding node local vxlan list into endpoint list seems not a good solution, because daemon cannot find out which are node ip addresses in endpoint list.

Can we keep the node local vxlan ip list an independent ip list? Just like copy the `node.Annotations[constants.AnnotationNodeLocalVxlanIPList]` into remote vtep CR's annotations.

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,219 @@
...
+				runtimeutil.HandleError(err)
+			}
+		}
+	}()
+
+	go func() {
+		defer wg.Done()
+		for _, v := range update {
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				vtep, err := m.LocalClusterRamaClient.NetworkingV1().RemoteVteps().Update(context.TODO(), v, metav1.UpdateOptions{})
```
hhyasdf, line 76: ditto.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,210 @@
...
+	if !newVtep && !endpointILListChanged && !vtepChanged {
+		return nil
+	}
+	if newVtep {
+		remoteVtep, err = m.LocalClusterRamaClient.NetworkingV1().RemoteVteps().Create(context.TODO(), remoteVtep, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+	} else {
+		remoteVtep, err = m.LocalClusterRamaClient.NetworkingV1().RemoteVteps().Update(context.TODO(), remoteVtep, metav1.UpdateOptions{})
```
hhyasdf, line 115: It seems remote vtep CRs will be updated both in node reconciling and ipinstance reconciling, will there be any conflict if the same remote vtep CR is updated at the same time? (There might be, because both of the two places are updating a whole remote vtep CR)

pkg/controller/remotecluster/cache.go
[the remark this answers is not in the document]
mars1024: sync.Map alos have method `Delete`

pkg/controller/remotecluster/cache.go
mars1024: I mean if someone have to shut down a cluster, close it firstly and delete it from cache secondly.
The two actions mentioned above do not have to be put into a single action.

pkg/controller/remotecluster/controller.go
mars1024: If not block here, the main thread will **force** clean all goroutines, which means more unexpected situations will happen.

pkg/controller/remotecluster/controller.go
mars1024: you mean `epoll` has no difference with `select`?

pkg/controller/remotecluster/controller.go
mars1024: do you know inline function?

pkg/controller/remotecluster/controller.go
Louhwz: how many networks can you have in one cluster?

pkg/controller/remotecluster/controller.go
Louhwz: yes, and then?

pkg/controller/remotecluster/rcmgr.go
Louhwz: I didn't do this because of the lock at first. If use the `Set()` function, the RWmutex will be locked twice and thus cause deadlock

pkg/controller/remotecluster/rcmgr.go
Louhwz: ditto

pkg/metrics/metrics.go
```diff
@@ -17,13 +17,17 @@
...
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 )
 
 const (
-	IPTotalUsageType     = "total"
-	IPUsedUsageType      = "used"
-	IPAvailableUsageType = "available"
+	IPTotalUsageType       = "total"
```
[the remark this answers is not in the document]
Louhwz: ok, i revert it

pkg/rcmanager/node.go
Louhwz: Now time `reconcileNode` always return nil after `diffNodeAndVtep`. How about using a global err to indicate that there are some error in `add` or `delete` or `update` goroutine and return it so if there are some conflicts, reconcile will happened.

pkg/rcmanager/ipinstance.go
Louhwz: ditto.

pkg/controller/remotecluster/rcmgr.go
Louhwz: If so, will there be some problem? For example, if `updateRemoteClusterStatus` is triggered by the timer when `addOrUpdateRCMgr` is called too.  The for loop in `updateRemoteClusterStatus`  will get a remote cluster manager, then `rcMgrCache.Del()` in `addOrUpdateRCMgr` is called and delete the old manager and recreate a new remote cluster manager.

hhyasdf: I don't think the same problem will not happen for current version of codes, because the mutex of `rcMgrCache` will not be locked any more either in `updateRemoteClusterStatus` after getting a remote cluster manager from `rcMgrCache`.

So maybe focus on the `c.rcMgrCache.RWMutex.Lock()` is not the right way to deal with your problem.

pkg/rcmanager/node.go
hhyasdf: I don't think what I said will cause an "error" in your code, but something strange will happen. Maybe both of the operations will success but one of them will seems haven't been executed.

Dockerfile.arm64
Louhwz: ditto.

pkg/controller/remotecluster/rcmgr.go
Louhwz: Yes, and so far I think that getting the old remote cluster manager is not a big problem. I prefer directly using locked function and we can remove the `RWMutex` and use `sync.Map` instead.

pkg/rcmanager/node.go
Louhwz: Do you mean one of the two reconciling may modify in an old replica because another has applied yet? If so, API server uses optimistic concurrency control to deal with this possible conflicts and thus will return an error from `Update` function. (Anyway, the `RetryOnConflict` is incorrect usage and is useless)

pkg/controller/remotecluster/rcmgr.go
hhyasdf: > Yes, and so far I think that getting the old remote cluster manager is not a big problem. I prefer directly using locked function and we can remove the `RWMutex` and use `sync.Map` instead.

OK, maybe just change to `sync.Map` then we can resolve this conversation.

pkg/rcmanager/node.go
hhyasdf: I'm not wondering if `Update` function is work correctly, but the remote vtep you used to update. 

If you have get the same remote vtep CR first, then you make some changes to them somehow before both Update operation. In ipinstance reconciling, you just update the `EndpointIPList`, while you update the `vtepIP` and `vtepIP` in node reconciling. 

Now we will find that the two remote vtep CRs you use to Update is different, only the last one being updated will become the result, while no conflict will happened.

hhyasdf: > I'm not wondering if `Update` function is work correctly, but the remote vtep you used to update.
> 
> If you have get the same remote vtep CR first, then you make some changes to them somehow before both Update operation. In ipinstance reconciling, you just update the `EndpointIPList`, while you update the `vtepIP` and `vtepIP` in node reconciling.
> 
> Now we will find that the two remote vtep CRs you use to Update is different, only the last one being updated will become the result, while no conflict will happened.

This will never happen because there is version checking before `Update` an exist object, if the version is changed, `Update` will not success.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,196 @@
...
+	endpointILListChanged = utils.DifferentSetFromStringSlice(remoteVtep.Spec.EndpointIPList, desired)
+
+	remoteVtep = remoteVtep.DeepCopy()
+	remoteVtep.Spec.VtepIP = vtepIP
+	remoteVtep.Spec.VtepMAC = vtepMac
+	remoteVtep.Spec.EndpointIPList = desired
+	remoteVtep.Status.LastModifyTime = curTime
+
+	vtepChanged = vtepIP != "" && vtepMac != "" && (vtepIP != remoteVtep.Spec.VtepIP || vtepMac != remoteVtep.Spec.VtepMAC)
+	if !newVtep && !endpointILListChanged && !vtepChanged {
```
hhyasdf, line 95: Maybe `node.Annotations[constants.AnnotationNodeLocalVxlanIPList]` changed should be count on `vtepChanged`, just link `vtepIP` and `vtepIP` are changed.

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		endpointIPList, err := m.pickEndpointListFromNode(node)
+		if err != nil {
+			continue
+		}
+		vtepIP := node.Annotations[constants.AnnotationNodeVtepIP]
+		vtepMac := node.Annotations[constants.AnnotationNodeVtepMac]
+		if vtep, exists := vtepMap[vtepName]; exists {
+			endpointIPListChanged := utils.DifferentSetFromStringSlice(endpointIPList, vtep.Spec.EndpointIPList)
+			if vtep.Spec.VtepIP != vtepIP || vtep.Spec.VtepMAC != vtepMac || endpointIPListChanged {
```
hhyasdf, line 133: ditto.

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,223 @@
...
+func (m *Manager) updateNode(oldObj, newObj interface{}) {
+	oldNode, _ := oldObj.(*apiv1.Node)
+	newNode, _ := newObj.(*apiv1.Node)
+	newNodeAnnotations := newNode.Annotations
+	oldNodeAnnotations := oldNode.Annotations
+
+	if newNodeAnnotations[constants.AnnotationNodeVtepIP] == "" || newNodeAnnotations[constants.AnnotationNodeVtepMac] == "" {
+		return
+	}
+	if newNodeAnnotations[constants.AnnotationNodeVtepIP] == oldNodeAnnotations[constants.AnnotationNodeVtepIP] &&
```
hhyasdf, line 214: ditto.

pkg/rcmanager/ipinstance.go
```diff
@@ -0,0 +1,196 @@
...
+		// vtepIP or vtepMac changed
+		vtepChanged           bool
+		endpointILListChanged bool
+	)
+
+	desired, err = m.pickEndpointListFromNode(node)
+	if err != nil {
+		return err
+	}
+	endpointILListChanged = utils.DifferentSetFromStringSlice(remoteVtep.Spec.EndpointIPList, desired)
```
hhyasdf, line 86: `endpointILListChanged` to `endpointListChanged`?

cmd/webhook/main.go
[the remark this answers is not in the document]
hhyasdf: "Informer in webhook" problem has another conversation, this will be set resolved.

pkg/controller/remotecluster/cache.go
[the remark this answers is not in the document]
hhyasdf: "sync.Map" problem has another conversation https://github.com/alibaba/hybridnet/pull/44#pullrequestreview-750012240, this will be set resolved.

pkg/controller/remotecluster/rcmgr.go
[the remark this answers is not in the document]
mars1024: Hi, bro, also think no need for this, one queue is enough for construction and bootstrap.

pkg/controller/remotecluster/controller.go
Louhwz: ok

cmd/webhook/main.go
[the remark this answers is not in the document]
hhyasdf: "Informer in webhook" is a later considering.

yamls/remotecluster-example.yaml
```diff
@@ -0,0 +1,10 @@
...
...SnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBTjFiCmZrTGFLT2RpcjhDQ2ZkTkF6VzV5aWRqRUVsL2VDRU5lNzVsRVdTUExvY2MxRlFyZHpiWUwrU0xBTTIxTUN0UG4KSW1BUW1jREEvODVaWTJhWVE0SDdxMXRBekFZcVJJRk55SmVpQnJReE5hUFl0SU5ldkhOYlFNLzk3RENYS2dFeAo0R3RkM01tcDBLWWZYaVVheE13SDZqSC9BRGREcmZQR0JnclZXTFNxaE1QTmZubGg1ZVd5WWcycWZ1WGRjWEk5CjAyV3RnMGh0dEtBbjZRTDlIOUtNUmlhd3kwYm1CWnFCRVJqa1QzN3JNc1pZVUd5UUsvKzRsbStsdlhEY3JaRkYKVHM5eVRFY3VTaG5xTHV5Zmp6RzNPUStMcUZVODhuL29VZlQ4VmNNM3B2ZUhGTUJNb1IrdHQ3VGNFSXRuZ1kwRgpQcFNhcWM2Um4wbVpYcTYyS1RzQ0F3RUFBYU5DTUVBd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0hRWURWUjBPQkJZRUZBb3pFSnV1dVRreG1HdlUrc2YzQ0xjd0piL0xNQTBHQ1NxR1NJYjMKRFFFQkN3VUFBNElCQVFBbnNlYUgwVW9kWU1uUjJtQ0lMTXNzdHRoVU9uRzRLcHA1b0Q2RVo2UGVVekh4WDVHbwpUWVJDWXpWSExIem9aVTdQV002aHJKUGh3YncrZnJtcXhCazJ2cnExc3hkVm11b0tqK3kwNTNMTmdsbDUyR2RTClV6VFp3YWVzc1dUMmtaaENYd0dpR01ES01ZWXJkWXAvVlVoc2lyZTJ5UjlacWc4elluaW4ra2xVWjQrZkhXZmQKU2lCc01XcGZ1dmZFSmdrVFViVGNIS1Jvc3NiZDBBR0hjcFdWUkN3aFRjME8wbTB0SXo4SVRxQTVGVk1nUmpSVgp1MEFCZXdnTHdmSHFKSlRTMG9waTJLcmFwcmc2aGRxVnBVdFBQUjlHY1BiMmRZTjc5ZC9lM3RqS2VSU2tpNUE4CkJyeXVRcmoyZElURk5xa0VZRXlJbTUzaHh0SXNhLzREVmFlcQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
mars1024, line 8: redact these parts about secrets

yamls/rama-init.yaml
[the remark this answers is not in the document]
mars1024: we can leave this for now, because it is changeable

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,231 @@
...
+			_ = m.LocalClusterRamaClient.NetworkingV1().RemoteVteps().Delete(context.TODO(), v, metav1.DeleteOptions{})
+			if err != nil && !k8serror.IsNotFound(err) {
+				errHappen = true
+				klog.Warningf("Can't delete remote vtep in local cluster. remote vtep name=%v", v)
+			}
+		}
+	}()
+
+	wg.Wait()
+	if errHappen {
```
mars1024, line 106: use channel to transfer errors between goroutines?

pkg/rcmanager/node.go
```diff
@@ -0,0 +1,231 @@
...
+		defer m.NodeQueue.Done(obj)
+		var (
+			key string
+			ok  bool
+		)
+		if key, ok = obj.(string); !ok {
+			m.NodeQueue.Forget(obj)
+			return nil
+		}
+		if err := m.reconcileNode(); err != nil {
```
mars1024, line 177: FIXME: use event to trigger node reconciliation

pkg/rcmanager/rcmanager.go
```diff
@@ -0,0 +1,233 @@
...
+		if ok := cache.WaitForCacheSync(managerCh, m.NodeSynced, m.SubnetSynced, m.IPSynced, m.NetworkSynced, m.RemoteClusterNodeSynced); !ok {
+			klog.Errorf("failed to wait for remote cluster caches to sync. clusterName=%v", m.ClusterName)
+			return
+		}
+		go wait.Until(m.RunNodeWorker, 1*time.Second, managerCh)
+		go wait.Until(m.RunSubnetWorker, 1*time.Second, managerCh)
+		go wait.Until(m.RunIPInstanceWorker, 1*time.Second, managerCh)
+	}()
+	go m.KubeInformerFactory.Start(managerCh)
+	go m.RamaInformerFactory.Start(managerCh)
```
mars1024, line 214: FIXME: how to know the status of manager?

mars1024 (approved):
/lgtm
/approve

PR #48, opened 2021-08-30
taking higher priority on specified-network annotation/label on selecting network for pod

mars1024:

Pull Request Description
---

When pod has a specified-network annotation which direct to a overlay network but has no network-type annotation, then IPAM controller will pick a underlay network which will cause conflicts with the overlay network above.
 
NONE

### Describe how you did it
Pick network for pod based on specified-network annotation/label firstly.

### Describe how to verify it
NONE

### Special notes for reviews
NONE

### DIFF
--- a/pkg/webhook/mutating/pod.go
+++ b/pkg/webhook/mutating/pod.go
@@ -111,12 +111,20 @@ func PodCreateMutation(ctx context.Context, req *admission.Request, handler *Han
 		}
 	}
 
-	var networkType = ipamtypes.ParseNetworkTypeFromString(utils.PickFirstNonEmptyString(pod.Annotations[constants.AnnotationNetworkType], pod.Labels[constants.LabelNetworkType]))
+	var networkTypeFromPod = utils.PickFirstNonEmptyString(pod.Annotations[constants.AnnotationNetworkType], pod.Labels[constants.LabelNetworkType])
+	var networkType = ipamtypes.ParseNetworkTypeFromString(networkTypeFromPod)
 	if len(networkName) > 0 {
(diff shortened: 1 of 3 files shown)


hhyasdf (commented):
Maybe https://github.com/alibaba/hybridnet/blob/main/pkg/webhook/mutating/pod.go#L114 should also be fixed. 

If a pod is specified to a overlay Network and the default network type is Underlay, pod should be mutated for patching nodeSeletor.

pkg/webhook/mutating/pod.go
```diff
@@ -111,22 +111,30 @@ func PodCreateMutation(ctx context.Context, req *admission.Request, handler *Han
...
+			case ipamtypes.Overlay:
+				// overlay network has a wide scheduling domain over the whole cluster
+				// no more selectors should be patched
+			default:
+				return admission.Errored(http.StatusBadRequest, fmt.Errorf("unknown network type %s", networkType))
+			}
+		} else {
+			// specified network takes higher priority than network type defaulting, if no network type specified
+			// from pod, then network type should inherit from network type of specified network from pod
+			networkType = ipamtypes.ParseNetworkTypeFromString(string(networkingv1.GetNetworkType(network)))
```
hhyasdf, line 137: If a pod is specified to an Underlay network while without specific network type, the steps of patching node selector seems to be lost here?

mars1024: good catch!

hhyasdf (approved):
/lgtm

hhyasdf (approved):

PR #49, opened 2021-08-31
bugfix: fix iptables masquerade local pod traffic error

hhyasdf:
Pull Request Description
---

Exist iptables rules will masquerade the traffic from remote pods to local pods, which is not supposed to be done.

NONE

### Describe how you did it
Ignore the traffic from remote pods to local pods by container host interface name. To do this, interface name "xxx_h" should be changed to "h_xxx", because iptables cannot recognize interfaces by suffix. 

### Describe how to verify it
Use tcpdump in overlay pod, the traffic from remote overlay pods to local pods should not be masqueraded anymore.

### Special notes for reviews
Only new pod's traffic with "h_xxx" interface will be fixed, old pod with "xxx_h" named host interface should be recreate to fix this issue.

### DIFF
--- a/pkg/daemon/containernetwork/constants.go
+++ b/pkg/daemon/containernetwork/constants.go
@@ -20,7 +20,7 @@ const (
 	DockerNetnsDir     = "/var/run/docker/netns"
 	ContainerdNetnsDir = "/var/run/netns/"
 
-	ContainerHostLinkSuffix = "_h"
+	ContainerHostLinkPrefix = "h_"
 	ContainerHostLinkMac    = "ee:ee:ee:ee:ee:ee"
 	ContainerInitLinkSuffix = "_c"
 	VxlanLinkInfix          = ".vxlan"
--- a/pkg/daemon/addr/addr.go
+++ b/pkg/daemon/addr/addr.go
@@ -19,7 +19,6 @@ package addr
 import (
 	"fmt"
 	"net"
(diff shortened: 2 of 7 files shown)


pkg/daemon/containernetwork/constants.go
```diff
@@ -20,7 +20,7 @@ const (
 	DockerNetnsDir     = "/var/run/docker/netns"
 	ContainerdNetnsDir = "/var/run/netns/"
 
-	ContainerHostLinkSuffix = "_h"
+	ContainerHostLinkPrefix = "h_"
```
mars1024, line 23: What if supporting both prefix and suffix?
1. Check both
2. Create prefix

Then the upgrading will become more smooth.

hhyasdf: Actually, "ContainerHostLinkPrefix" is only used to generate veth pair name, if in a new environment with newest hybridnet version, "ContainerHostLinkSuffix" is not needed anymore. 

While upgrading, "_h" is also checked only one place of all the codes, so we don't need to keep the "ContainerHostLinkSuffix" variable for upgrading.

hhyasdf: Unlike "ip-pools" annotation, this will not be used directly by users.

mars1024 (approved):
/lgtm
/approve

PR #79, opened 2021-09-26
webhook: discard informer cache in webhook, skip uniqueness check of self and some typos

mars1024:
Pull Request Description
---

1. Discard informer cache in webhook, btw, signal
2. Skip uniqueness check of self
3. Some comments and typos

NONE

### Describe how you did it
NONE

### Describe how to verify it
NONE

### Special notes for reviews
NONE


pkg/webhook/validating/remotecluster.go
```diff
@@ -44,57 +43,57 @@ func init() {
...
-	if err != nil {
+	// ensure the uniqueness of cluster config
+	rcs := &ramav1.RemoteClusterList{}
+	if err = handler.Client.List(ctx, rcs); err != nil {
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
-	for _, rc := range rcs {
-		if uuid == rc.Status.UUID {
-			return admission.Denied("Duplicate cluster configuration")
+	for i := range rcs.Items {
```
hhyasdf, line 90: Maybe use `for _, item := range rcs.Items` to get the items in an array directly is better instead of using index?

mars1024: items is an array of value type, so use index will save cost of value copy.

hhyasdf: Copy that.

hhyasdf (approved):
/lgtm

PR #80, opened 2021-09-26
controller: use thread-safe map instead of self-built cache

mars1024:
Pull Request Description
---

1. Self-built cache is more like a built-in `sync.Map`
2. Close and deletion should not be compressed into a single method

NONE

### Describe how you did it
NONE

### Describe how to verify it
NONE

### Special notes for reviews
NONE


hhyasdf (approved):
/lgtm

PR #108, opened 2021-11-09
bugfix: reduce cpu cost for large scale

hhyasdf:
Pull Request Description
---

For a large scale environment, daemon will cost a lot of cpu usage because of a large number of neigh link events.

NONE.

### Describe how you did it
Increase netlink channel size to avoid blocks while receiving and sending netlink update events.

### Describe how to verify it
Run daemon in a a large scale environment.

### Special notes for reviews

### DIFF
--- a/cmd/daemon/daemon.go
+++ b/cmd/daemon/daemon.go
@@ -17,8 +17,6 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


mars1024 (approved):
/lgtm
Thanks!

PR #111, opened 2021-11-12
fix arp proxy exit in large scale cluster

hhyasdf:
Pull Request Description
---

In a large scale cluster, the number of vxlan neigh cache entries on the every node might grow to a huge size, which will bring a "No buffer space available" failure for netlink.NeighSubscribe function. Once NeighSubscribe function is exit, userspace arp proxy will never start again. (The NeighSubscribe function will also close the event channel passed to it.)

NONE.

### Describe how you did it
In this PR, I fix the NeighSubscribe exist problem and change the behaviors of arp proxy to control the size of neigh cache size:
1. Multicast arp/ndp request will only be sent if a neigh cache is `INCOMPLETE`, in other words, when neigh cache is just created. So we just need to handle the `INCOMPLETE` netlink neigh event, and proxy arp for it.
2. Unicast arp/ndp probe request will be sent if a neigh cache's state is changed from `STALE` to `PROBE`. This will not be handled in arp proxy, so every neigh cache can be `STALE`. Every pod will generate an arp unicast probe for every 5 second. Because `unicast_solicit` has a higher priority than `app_solicit`, we cannot handle the netlink events of `NUD_DELAY` or `NUD_PROBE` before the unicast arp probe packet is send out.
3. ~~Gratuitous arp/ndp packets will be sent on overlay forward interface once, while an overlay/underlay pod is being created.~~
4. Daemon will clean STALE and FAILED neigh caches periodically on vxlan interface, default once an hour.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -63,6 +63,8 @@ const (
 	NeighUpdateChanSize = 2000
 	LinkUpdateChainSize = 200
 	AddrUpdateChainSize = 200
+
+	NetlinkSubscribeRetryInterval = 10 * time.Second
 )
 
 // Controller is a set of kubernetes controllers
@@ -213,10 +215,13 @@ func NewController(config *daemonconfig.Configuration,
 		UpdateFunc: controller.enqueueUpdateSubnet,
 	})
 
-	ipInstanceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    controller.enqueueAddOrDeleteIPInstance,
-		UpdateFunc: controller.enqueueUpdateIPInstance,
-		DeleteFunc: controller.enqueueAddOrDeleteIPInstance,
+	ipInstanceInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: controller.filterIPInstance,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    controller.enqueueAddOrDeleteIPInstance,
+			UpdateFunc: controller.enqueueUpdateIPInstance,
+			DeleteFunc: controller.enqueueAddOrDeleteIPInstance,
+		},
 	})
 
 	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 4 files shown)


pkg/daemon/controller/controller.go
```diff
@@ -316,33 +321,83 @@ func (c *Controller) GetIPInstanceSynced() cache.InformerSynced {
 //
 // Restart of vxlan interface will also trigger subnet and ip instance reconcile loop.
 func (c *Controller) handleLocalNetworkDeviceEvent() error {
+	hostNetNs, err := netns.Get()
+	if err != nil {
+		return fmt.Errorf("get root netns failed: %v", err)
+	}
+
 	linkCh := make(chan netlink.LinkUpdate, LinkUpdateChainSize)
```
mars1024, line 329: If `Subscribe` is created in loop, then L329-L332 is duplicated.

pkg/daemon/controller/controller.go
```diff
@@ -425,49 +480,65 @@ func (c *Controller) handleVxlanInterfaceNeighEvent() error {
...
+	timer := time.NewTicker(c.config.VxlanExpiredNeighCachesClearInterval)
+	errorMessageWrapper := initErrorMessageWrapper("handle vxlan interface neigh event failed: ")
 
 	go func() {
-		errorMessageWrapper := initErrorMessageWrapper("handle vxlan interface neigh event failed: ")
+		for {
+			// Clear stale and failed neigh entries for vxlan interface at the first time.
+			// Once neigh subscribe failed, it neigh fail because of "No buffer space available" problem,
+			// need to clean expired neigh caches
+			if err := clearVxlanExpiredNeighCaches(); err != nil {
```
mars1024, line 496: caches clear may not be in a loop

pkg/daemon/controller/ipinstance.go
```diff
@@ -37,6 +37,16 @@ import (
...
 )
 
+func (c *Controller) filterIPInstance(obj interface{}) bool {
+	p, ok := obj.(*networkingv1.IPInstance)
+	if !ok {
+		return false
+	}
+
+	// only ip on this node
+	return p.Labels[constants.LabelNode] == c.config.NodeName
```
mars1024, line 47: should empty string be ignored here?

pkg/daemon/controller/controller.go
hhyasdf: Yes I forgot this. Fixed.

pkg/daemon/controller/ipinstance.go
hhyasdf: Fixed. We should not ignore it.

pkg/daemon/controller/controller.go
hhyasdf: As it says in comments, the goroutine of `netlink.NeighSubscribe` might exit most probably because of neigh caches buffer space error. In this situation, we are supposed to clean the expired neigh caches once in case it fails over and over again, while normally it only work once if the goroutine of `netlink.NeighSubscribe` never exit.

mars1024: got it!

pkg/daemon/controller/ipinstance.go
```diff
@@ -37,6 +37,18 @@ import (
...
+func (c *Controller) filterIPInstance(obj interface{}) bool {
+	p, ok := obj.(*networkingv1.IPInstance)
+	if !ok {
+		return false
+	}
+
+	nodeLabel := p.Labels[constants.LabelNode]
+
+	// only ip on this node
+	return len(nodeLabel) == 0 || nodeLabel == c.config.NodeName
```
mars1024, line 49: Only IPInstances with valid node should be processed.

hhyasdf: I thought you mean that a node label of an ip instance could be empty xD.  What did the "should empty string be ignored here?" means in the old conversation?  Maybe you want `len(nodeLabel) != 0 && nodeLabel == c.config.NodeName`?

mars1024 (approved):
/lgtm
/approve
Thanks!

PR #118, opened 2021-11-17
bugfix: fix ipv6 addr calculate error

hhyasdf:
Pull Request Description
---

Fix ipv6 address calculating problems, two functions will fail if use ipv6 subnet.

NONE.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #128, opened 2021-12-01
allow specified subnets without networks

mars1024:
Pull Request Description
---

If subnet specified, then network should be validated and auto-specified.

### Special notes for reviews

### DIFF
--- a/pkg/webhook/validating/pod.go
+++ b/pkg/webhook/validating/pod.go
@@ -22,17 +22,17 @@ import (
 	"net/http"
 	"strings"
 
-	networkingv1 "github.com/alibaba/hybridnet/pkg/apis/networking/v1"
-	"github.com/alibaba/hybridnet/pkg/constants"
-	"github.com/alibaba/hybridnet/pkg/feature"
-	ipamtypes "github.com/alibaba/hybridnet/pkg/ipam/types"
-	"github.com/alibaba/hybridnet/pkg/utils"
-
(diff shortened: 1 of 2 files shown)


pkg/webhook/validating/pod.go
```diff
@@ -72,7 +72,34 @@ func PodCreateValidation(ctx context.Context, req *admission.Request, handler *H
...
+				))
+			}
+		} else {
+			// subnet can also determine the specified network
+			specifiedNetwork = subnet.Spec.Network
+		}
+	}
 
+	// Existing IP Instances Validation
+	if len(specifiedNetwork) > 0 {
```
hhyasdf, line 102: Do we need to check network type again here? If there is specified subnet only.

mars1024: network not network-type here

mars1024: If a subnet is specified, then the network will be determined by this. Of course, we should check potential conflicts of network and existing IP instances.

hhyasdf: > network not network-type here

I mean, this line is fine. But if the network and network type of pod are not specified manually, `networkType`  will be empty here even if we have a specified subnet. Maybe `networkType` should be decided again above this line.

hhyasdf (approved):
/lgtm

hhyasdf:
Finish #143

PR #134, opened 2022-01-18
Daemon controller-runtime adapt

hhyasdf:
Pull Request Description
---

### Special notes for reviews


pkg/daemon/server/handle.go
```diff
@@ -97,9 +89,9 @@ func (cdh cniDaemonHandler) handleAdd(req *restful.Request, resp *restful.Respon
 		time.Sleep(backOffBase)
 		backOffBase = backOffBase * 2
 
-		pod, err := cdh.KubeClient.CoreV1().Pods(podRequest.PodNamespace).Get(context.TODO(), podRequest.PodName, metav1.GetOptions{})
-		if err != nil {
-			errMsg := fmt.Errorf("get pod %v/%v failed: %v", podRequest.PodName, podRequest.PodNamespace, err)
+		pod := &corev1.Pod{}
+		if err := cdh.mgrAPIReader.Get(context.TODO(), types.NamespacedName{Name: podRequest.PodName}, pod); err != nil {
```
mars1024, line 93: why use APIReader not client here>

hhyasdf: I think `Client` will read from the cache in local memory while `APIReader` will not. And in order to get `Pod` information from apiserver, maybe using `APIReader` is necessary here.

PR #138, opened 2022-01-20
fix daemon controller runtime error

hhyasdf:
Pull Request Description
---

### Special notes for reviews

### DIFF
--- a/cmd/daemon/daemon.go
+++ b/cmd/daemon/daemon.go
@@ -48,6 +48,7 @@ func main() {
 		entryLog.Error(err, "failed to parse config")
 		os.Exit(1)
 	}
+	entryLog.Info("generate daemon config", "config", *config)
 
 	// setup manager
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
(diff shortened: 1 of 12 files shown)


cmd/daemon/daemon.go
```diff
@@ -43,7 +43,7 @@ func main() {
 	entryLog.Info("starting hybridnet daemon",
 		"known-features", feature.KnownFeatures(), "commit-id", gitCommit)
 
-	config, err := daemonconfig.ParseFlags()
+	config, err := daemonconfig.ParseFlags(entryLog)
```
mars1024, line 46: no need to take logger as an input, we can log error and config outside

hhyasdf: Yes, I've found that too. Fixed.

pkg/daemon/controller/utils.go
```diff
@@ -55,7 +58,10 @@ func (t *simpleTriggerSource) Start(ctx context.Context, handler handler.EventHa
 }
 
 func (t *simpleTriggerSource) Trigger() {
-	t.queue.Add(t.key)
+	t.queue.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		Name:      t.key,
+		Namespace: "All",
```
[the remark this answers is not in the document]
hhyasdf: Fixed.

mars1024 (approved):
/lgtm

PR #153, opened 2022-02-14
bugfix: pod network selector not patched if only network specified

hhyasdf:
Pull Request Description
---

Fix the problem that pod network selector is not patched if only network specified without a specified subnet.

### Special notes for reviews


mars1024 (approved):
/lgtm
/approve
Thanks!

PR #159, opened 2022-02-16
zap flags and a bugfix

mars1024:
Pull Request Description
---

1. use flag to set zap logging options
2. fix bug on pod controller enqueueing

### Special notes for reviews


hhyasdf (approved):
/lgtm

PR #163, opened 2022-02-18
fix the error of new nodes' vtep info not being added to local vxlan dev fdb

hhyasdf:
Pull Request Description
---

1. Update some docs.
2. Fix the error of new nodes' vtep info not being added to local vxlan dev fdb.
3. Avoid unneeded error output for bgp manager.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #165, opened 2022-02-21
controller: fix uninitialized nil map

mars1024:

Pull Request Description
---

Type `uuidMutex` has uninitialized map field, try to fix this bug.

### Special notes for reviews


hhyasdf (approved):
/lgtm

PR #166, opened 2022-02-22
zap log defaulting and bugfix on condition

mars1024:
Pull Request Description
---

1. defaulting log encoder to console mode
2. explicit reason in condition to avoid status update failing

### Special notes for reviews


hhyasdf (approved):
/lgtm

PR #176, opened 2022-02-28
manager-runtime: refactor daemon run cause manager does not support restart

mars1024:

Pull Request Description
---

`Manager` of controller-runtime does not support restarting for now, so if manager exits unexpectedly, daemon should not call `manager.Start` again and again.
Instead, in restart progress, another new manager should be constructed and initialized.

### Special notes for reviews


hhyasdf (approved):
/lgtm

PR #184, opened 2022-03-03
improve: no new pod gets running if all the bgp peers are down

hhyasdf:
Pull Request Description
---

### Special notes for reviews

### DIFF
--- a/pkg/daemon/containernetwork/utils.go
+++ b/pkg/daemon/containernetwork/utils.go
@@ -154,24 +154,25 @@ func checkPodNetConfigReady(podIP net.IP, podCidr *net.IPNet, forwardNodeIfIndex
 			}
 
 			defaultRouteExist := false
-			bgpPathExist := false
-
 			if ruleExist {
 				defaultRouteExist, err = daemonutils.CheckDefaultRouteExist(table, family)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/daemon/containernetwork/utils.go
```diff
@@ -152,26 +152,22 @@ func checkPodNetConfigReady(podIP net.IP, podCidr *net.IPNet, forwardNodeIfIndex
 			if err != nil {
 				return fmt.Errorf("failed to check cidr %v rule and default route exist: %v", podCidr, err)
 			}
+			defaultRouteExist, err := daemonutils.CheckDefaultRouteExist(table, family)
```
mars1024, line 155: Why change the dependencies of these checks? I mean, if the first check failed, there is no need to check the others.

hhyasdf: Because there will be 4 layers of nested `if` block, which seems not clean enough. And it's fine to always run all the checks, none of them will fail with an error easily.

hhyasdf: Fixed. 

For the rule check and default route check, they do have a dependency. But for the others, they have no dependencies and will not fail with an error easily. I do this to reduce the number of nested `if` blocks.

mars1024 (approved):
/lgtm

Issue #121, opened 2021-11-18
Using charts for user-friendly deployment


mars1024:
Issue Description
---

Type: *feature request*

### Describe what feature you want
Use charts to deploy hybridnet, including
- versioning
- custom parameters
- one-command installation

hhyasdf:
Thanks for your advice, now you can find hybridnet charts in https://artifacthub.io/packages/search?repo=hybridnet.

PR #199, opened 2022-03-15
chore: init helm charts

hhyasdf:
Pull Request Description
---

Use helm chart to provide a more friendly way of installation.

Deal with #121 .

### Special notes for reviews


charts/hybridnet/templates/services.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
mars1024, line 1: seems useless two lines

charts/hybridnet/templates/networkinitialisation.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
mars1024, line 1: typo in file name

charts/hybridnet/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+  hybridnet:
+    image: hybridnetdev/hybridnet
+    tag: v0.4.1
+  registryURL: "docker.io"
+
+# -- It's always encouraged to use an overlay network to finish some general jobs in hybridnet. This
+# chart will always create an "init" overlay Network/Subnet to initialize the K8s cluster.
+initNetworkConfig:
+  # -- CIDR of "init" Subnet.
+  cidr: 100.64.0.0/16
```
mars1024, line 11: why not use a private subnet?

charts/hybridnet/templates/networkinitialisation.yaml
hhyasdf: fixed

charts/hybridnet/values.yaml
hhyasdf: Because it's not friendly. Normally, if hybridnet is installed in a fully free-downloaded K8s cluster, pod will not have a subnet-specified annotation. If the init Subnet is private, extra operations need to be finished before installing hybridnet, which brings some wasted time.

charts/hybridnet/templates/services.yaml
hhyasdf: fixed

mars1024 (approved):
/lgtm
/approve
Thanks!

PR #200, opened 2022-03-17
bugfix: static ip migrating error

hhyasdf:
Pull Request Description
---

If a sts pod with a static ip is recreate and rescheduled from one node to another, it will fail to be created. Because daemon has ignored such IPInstance "Update" events

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/ipinstance.go
+++ b/pkg/daemon/controller/ipinstance.go
@@ -122,6 +122,11 @@ func (r *ipInstanceReconciler) Reconcile(ctx context.Context, request reconcile.
 	r.ctrlHubRef.bgpManager.ResetIPInfos()
 
 	for _, ipInstance := range ipInstanceList.Items {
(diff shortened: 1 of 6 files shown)


pkg/daemon/controller/ipinstance.go
```diff
@@ -209,8 +214,6 @@ func (r *ipInstanceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	if err := ipInstanceController.Watch(&source.Kind{Type: &networkingv1.IPInstance{}},
 		&fixedKeyHandler{key: ActionReconcileIPInstance},
-		&predicate.ResourceVersionChangedPredicate{},
```
mars1024: why remove this two predicates?

hhyasdf: Fixed. I had a mistake with `resource version` and `generation`.

mars1024 (approved):
/lgtm

PR #208, opened 2022-03-24
feat: support specify mac addr via annotation

### LINKED ISSUE #190: support MAC assignment through pod annotation
Issue Description
---

Type: *feature request*

### Describe what feature you want
1. assign MAC address in pod annotation
2. check MAC conflict

sjtufl:
Pull Request Description
---

Introduce a new feature: support specifying MAC address via pod annotation.

Fixes #190 

### Describe how you did it

1. Verify the validness of specified MAC address in validating webhook;
2. Check whether the specified MAC address is in conflict with existing IP instances.

### Describe how to verify it
Specify a static MAC address via annotation `"networking.alibaba.com/specified-mac"` on Pod. 

### Special notes for reviews

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -23,15 +23,12 @@ import (
 	"os"
 	"time"
 
-	globalutils "github.com/alibaba/hybridnet/pkg/utils"
-
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	multiclusterv1 "github.com/alibaba/hybridnet/pkg/apis/multicluster/v1"
@@ -42,7 +39,9 @@ import (
 	"github.com/alibaba/hybridnet/pkg/controllers/networking"
 	"github.com/alibaba/hybridnet/pkg/controllers/utils"
 	"github.com/alibaba/hybridnet/pkg/feature"
+	"github.com/alibaba/hybridnet/pkg/ipam/store"
 	"github.com/alibaba/hybridnet/pkg/managerruntime"
+	globalutils "github.com/alibaba/hybridnet/pkg/utils"
 	zapinit "github.com/alibaba/hybridnet/pkg/zap"
 )
 
@@ -288,6 +287,18 @@ func initIndexers(mgr ctrl.Manager) (err error) {
 			default:
 				return nil
 			}
+		}); err != nil {
+		return err
+	}
+
+	// Init mac indexer for IPInstances
+	if err = mgr.GetFieldIndexer().IndexField(context.TODO(), &networkingv1.IPInstance{},
+		store.IndexerFieldMac, func(obj client.Object) []string {
+			ipInstance, ok := obj.(*networkingv1.IPInstance)
+			if !ok {
+				return nil
+			}
+			return []string{ipInstance.Spec.Address.MAC}
 		}); err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


hhyasdf:
What's the new feature for? I'm asking this because pod MAC address is not "exported" for hybridnet, in another word, a pod will always use the MAC address of its local node to communicate with the pods on other nodes.

sjtufl:
> What's the new feature for? I'm asking this because pod MAC address is not "exported" for hybridnet, in another word, a pod will always use the MAC address of its local node to communicate with the pods on other nodes. The same as CNI ipvlan.

Yes, it's true that user-specified MAC address is not necessary in terms of communicating with pods on other nodes. But we've also seen the need in some "non-communication" scenarios, e.g., internal auditing or generating product license. So I think it's more like a nice-to-have feature with no interference in core data-plane functionality.

hhyasdf:
Could you please provide a sample usage for this feature in "Describe how to verify it" section? 

I've noticed that a pod's MAC address can be specified with an `"networking.alibaba.com/specified-mac"` annotation. What will happen if the pod is create as a part of deployment? One specified MAC address for a pod seems not work out. 

Maybe it should be implemented similarly as `"networking.alibaba.com/ip-pool"`, which we already have.

pkg/controllers/networking/pod_controller.go
```diff
@@ -123,6 +124,25 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result
 		return ctrl.Result{}, nil
 	}
 
+	// It is not allowed that specified mac address is in conflict with existing IPInstances.
```
mars1024, line 127: can you check this in `createIP` of `IPAMStore`?

pkg/controllers/networking/pod_controller.go
```diff
@@ -536,6 +556,17 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) (err error) {
 		return err
 	}
 
+	// Init mac indexer for IPInstances.
+	if err = mgr.GetFieldIndexer().IndexField(context.TODO(), &networkingv1.IPInstance{}, indexerFieldMac, func(obj client.Object) []string {
```
mars1024, line 560: After the latest commit, this indexer registration may be moved to another place.

sjtufl:
> Could you please provide a sample usage for this feature in "Describe how to verify it" section?

OK, edited!

> I've noticed that a pod's MAC address can be specified with an `"networking.alibaba.com/specified-mac"` annotation. What will happen if the pod is create as a part of Deployment or StatefulSet? One specified MAC address for a pod seems not work out.
> 
> Maybe it should be implemented similarly as the annotation of `"networking.alibaba.com/ip-pool"`, which we already have in [wiki docs](https://github.com/alibaba/hybridnet/wiki/Static-pod-ip-addresses-for-StatefulSet).

Current implementation is only for single pod scenario and does not support workloads like Deployment or StatefulSet. Unlike IP retain, specifying static MAC address shouldn't be limited to stateful workloads. I think that specifying a "MAC-Pool" would be great, and it will require some additional work.

pkg/controllers/networking/pod_controller.go
sjtufl: Okay! It makes conflicts checked in a "global" manner.

hhyasdf:
> > Could you please provide a sample usage for this feature in "Describe how to verify it" section?
> 
> OK, edited!
> 
> > I've noticed that a pod's MAC address can be specified with an `"networking.alibaba.com/specified-mac"` annotation. What will happen if the pod is create as a part of Deployment or StatefulSet? One specified MAC address for a pod seems not work out.
> > Maybe it should be implemented similarly as the annotation of `"networking.alibaba.com/ip-pool"`, which we already have in [wiki docs](https://github.com/alibaba/hybridnet/wiki/Static-pod-ip-addresses-for-StatefulSet).
> 
> Current implementation is only for single pod scenario and does not support workloads like Deployment or StatefulSet. Unlike IP retain, specifying static MAC address shouldn't be limited to stateful workloads. I think that specifying a "MAC-Pool" would be great, and it will require some additional work.

Fine. If only for a single pod, maybe this should be considered as a feature for developers (of k8s controllers) but not for all the users. Looks good for me, cc @mars1024 .

hhyasdf:
Please rebase `main` branch to fix the golangci-lint error.

sjtufl:
This PR is closed since IPAM module is refactored a lot and leads to multiple conflicts. I will open another PR on supporting mac address pool assignment.

PR #221, opened 2022-04-15
remove checks for kubernetes service

hhyasdf:
Pull Request Description
---

The kubernetes svc check commands in policy.sh cannot work in a ipv6 only environment. Just remove it.

### Special notes for reviews


mars1024 (commented):
what's ipv6 only environment?

hhyasdf:
> 

Only ipv6 address on node, but pods will be DualStack.

hhyasdf:
> what's ipv6 only environment?

Only ipv6 address on node, but pods will be DualStack.

mars1024 (approved):
/lgtm

PR #225, opened 2022-04-15
remove label-changed event filter for daemon node controller

hhyasdf:
Pull Request Description
---

Node annotation changed should trigger the reconcile of daemon Node controller.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/node.go
+++ b/pkg/daemon/controller/node.go
@@ -304,7 +304,7 @@ func (r *nodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if err := nodeController.Watch(&source.Kind{Type: &corev1.Node{}},
 		&fixedKeyHandler{key: ActionReconcileNode},
 		&predicate.ResourceVersionChangedPredicate{},
-		&predicate.LabelChangedPredicate{},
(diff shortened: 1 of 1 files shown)


pkg/daemon/controller/node.go
```diff
@@ -304,7 +304,6 @@ func (r *nodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if err := nodeController.Watch(&source.Kind{Type: &corev1.Node{}},
 		&fixedKeyHandler{key: ActionReconcileNode},
 		&predicate.ResourceVersionChangedPredicate{},
-		&predicate.LabelChangedPredicate{},
```
mars1024, line 307: if all we needed is in annotations, how about replacing LabelChangedPredicate with AnnotationChangePredicate?

hhyasdf: There already has one, https://github.com/alibaba/hybridnet/pull/225/files#diff-bc1d985df6874b40c6a27274b7d19089f4624ce9d94d481cb60f44d7be9047b2L309.

hhyasdf: Fixed

mars1024 (approved):
/lgtm

PR #226, opened 2022-04-15
improve daemon vtep address selection

hhyasdf:
Pull Request Description
---

1. Make the vtep address of node can be changed easier.
2. Add `vtep-address-cidrs` flag for daemon to help select vtep address.

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/node.go
+++ b/pkg/daemon/controller/node.go
@@ -110,17 +110,6 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, request reconcile.Reques
 			r.ctrlHubRef.config.NodeName, err)
 	}
 
-	preVtepIP, exist := thisNode.Annotations[constants.AnnotationNodeVtepIP]
-	if exist {
-		// if vtep ip has been set and still exist on parent interface
-		for _, address := range existParentAddrList {
-			if address.IP.String() == preVtepIP {
-				vtepIP = address.IP
-				break
-			}
-		}
-	}
-
 	if vtepIP == nil {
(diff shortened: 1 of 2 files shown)


pkg/daemon/controller/node.go
```diff
@@ -78,63 +78,152 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, request reconcile.Reques
...
+	if err := ensureVxlanInterfaceAddresses(vxlanDev, nodeLocalVxlanAddrs); err != nil {
+		return reconcile.Result{Requeue: true}, fmt.Errorf("failed to ensure addresses for vxlan device %v: %v",
+			vxlanLinkName, err)
+	}
+
+	for _, node := range nodeList.Items {
+		if node.Annotations[constants.AnnotationNodeVtepMac] == "" ||
+			node.Annotations[constants.AnnotationNodeVtepIP] == "" ||
+			node.Annotations[constants.AnnotationNodeLocalVxlanIPList] == "" {
+			return reconcile.Result{Requeue: true}, fmt.Errorf("node %v's vtep information has not been updated", node.Name)
```
mars1024, line 127: is it better to skip this and aggregate all errors into one?

hhyasdf: If it's skipped, the logs might be flushed and make it harder to find the error. For the no-skipped version, there are two cases:
1. An error happens once start. The whole cluster will not be ready until error is fixed.
2. An error happens after something is running. New Pods will never get ready but exist Pods will not be effected.

I think that neither of the cases will bring accidents but only more efficiency of resolving problems.

mars1024 (approved):
/lgtm

PR #227, opened 2022-04-19
fix daemon ipv6 throw routes check failed error

hhyasdf:
Pull Request Description
---

Fix "to overlay subnet route table 40000 is used by others" error of daemon. It happens if an ipv6 subnet with excluded ip ranges is created.

### Special notes for reviews


mars1024 (approved):
/lgtm
/approve
Thanks！

PR #233, opened 2022-04-25
add flag to disable enhanced address

hhyasdf:
Pull Request Description
---

### Special notes for reviews

### DIFF
--- a/pkg/daemon/config/config.go
+++ b/pkg/daemon/config/config.go
@@ -93,6 +93,8 @@ type Configuration struct {
 	NeighGCThresh1 int
 	NeighGCThresh2 int
 	NeighGCThresh3 int
+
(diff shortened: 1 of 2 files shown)


pkg/daemon/config/config.go
```diff
@@ -93,6 +93,8 @@ type Configuration struct {
 	NeighGCThresh1 int
 	NeighGCThresh2 int
 	NeighGCThresh3 int
+
+	EnableVlanArpEnhancement bool
```
mars1024, line 97: I thinks this enhancement is working for both vlan and bridge.
How about rename this to `EnableArpEnhancedAddress`?

mars1024 (approved):
/lgtm

PR #235, opened 2022-04-26
Prevent enhanced addresses from source selection

### LINKED ISSUE #232: enhanced address is being used from node to local pods unexpectedlly
Bug Report
---

Type: *bug report*

### What happened

image

image

Enhanced address is being used, when we want to ping local pods from node. This will cause the ICMP reply will never come back.

### Environment
- hybridnet version: 3.2.0
- Kernel (e.g. `uname -a`):
- Kubernetes version:
- Install tools:
- Others:

hhyasdf:
Pull Request Description
---

Fix the error that arp enhanced addresses will be taken as source IP address by mistake.

Fix #232 

### Describe how you did it
Create arp enhanced addresses with scope `link`

### Special notes for reviews


mars1024 (approved):
/lgtm

Issue #219, opened 2022-04-13
Global BGP network support 


hhyasdf:
Issue Description
---

Type: *feature request*

### Describe what feature you want
Now we need to create one BGP mode hybridnet Network for each ToR. It can not satisfy the Pods which need both static IP addresses and cluster-wide scheduled ability.

### Additional context

It's possible to achieve it for BGP protocol, all we need to do is to change the way of advertising routes.

PR #240, opened 2022-05-06
introduce global bgp network

hhyasdf:
Pull Request Description
---

Introduce GlobalBGP Network to provide a cluster-scope BGP IPAM, which can only work with exist BGP Networks.

For #219 

### Special notes for reviews

### DIFF
--- a/pkg/apis/networking/v1/types.go
+++ b/pkg/apis/networking/v1/types.go
@@ -19,16 +19,18 @@ package v1
 type NetworkType string
 
 const (
-	NetworkTypeUnderlay = NetworkType("Underlay")
-	NetworkTypeOverlay  = NetworkType("Overlay")
+	NetworkTypeUnderlay  = NetworkType("Underlay")
+	NetworkTypeOverlay   = NetworkType("Overlay")
+	NetworkTypeGlobalBGP = NetworkType("GlobalBGP")
 )
 
 type NetworkMode string
 
 const (
-	NetworkModeBGP   = NetworkMode("BGP")
-	NetworkModeVlan  = NetworkMode("VLAN")
-	NetworkModeVxlan = NetworkMode("VXLAN")
+	NetworkModeBGP       = NetworkMode("BGP")
+	NetworkModeVlan      = NetworkMode("VLAN")
+	NetworkModeVxlan     = NetworkMode("VXLAN")
+	NetworkModeGlobalBGP = NetworkMode("GlobalBGP")
 )
 
 type Count struct {
--- a/pkg/apis/networking/v1/utils.go
+++ b/pkg/apis/networking/v1/utils.go
@@ -67,9 +67,21 @@ func GetNetworkMode(networkObj *Network) NetworkMode {
 		if len(networkObj.Spec.Mode) == 0 {
 			return NetworkModeVxlan
 		}
+	case NetworkTypeGlobalBGP:
+		if len(networkObj.Spec.Mode) == 0 {
+			return NetworkModeGlobalBGP
+		}
 	}
 
 	return networkObj.Spec.Mode
+}
+
+func IsGlobalNetwork(networkObj *Network) bool {
+	return IsGlobalNetworkType(GetNetworkType(networkObj))
+}
+
+func IsGlobalNetworkType(networkType NetworkType) bool {
+	return networkType == NetworkTypeOverlay || networkType == NetworkTypeGlobalBGP
 }
 
 func IsIPv6IPInstance(ip *IPInstance) bool {
--- a/pkg/apis/networking/v1/utils.go
+++ b/pkg/apis/networking/v1/utils.go
@@ -76,11 +76,11 @@ func GetNetworkMode(networkObj *Network) NetworkMode {
 	return networkObj.Spec.Mode
 }
 
-func IsGlobalNetwork(networkObj *Network) bool {
-	return IsGlobalNetworkType(GetNetworkType(networkObj))
+func IsGlobalUniqueNetwork(networkObj *Network) bool {
+	return IsGlobalUniqueNetworkType(GetNetworkType(networkObj))
 }
 
-func IsGlobalNetworkType(networkType NetworkType) bool {
+func IsGlobalUniqueNetworkType(networkType NetworkType) bool {
 	return networkType == NetworkTypeOverlay || networkType == NetworkTypeGlobalBGP
 }
 
--- a/pkg/controllers/networking/node_controller.go
+++ b/pkg/controllers/networking/node_controller.go
@@ -69,23 +69,43 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	var underlayAttached, overlayAttached bool
+	var underlayAttached, overlayAttached, bgpAttached bool
 	if underlayAttached, overlayAttached, err = utils.DetectNetworkAttachmentOfNode(r, node); err != nil {
 		log.Error(err, "unable to detect network attachment")
 		return ctrl.Result{}, err
 	}
 
-	nodePatch := client.MergeFrom(node.DeepCopy())
+	if underlayAttached {
+		if globalBGPNetwork, err := utils.FindGlobalBGPNetwork(r); err != nil {
+			log.Error(err, "unable to detect global bgp network exist")
+			return ctrl.Result{}, err
+		} else if len(globalBGPNetwork) != 0 {
+			// if global bgp network not exist, do not patch bgp attachment label to node
+			var bgpNetworkName string
+			if bgpNetworkName, err = utils.FindBGPUnderlayNetworkForNode(r, node.GetLabels()); err != nil {
+				log.Error(err, "unable to find bgp underlay network for node %v", node.Name)
+				return ctrl.Result{}, err
+			}
 
-	attachedToString := func(attached bool) string {
-		if attached {
-			return constants.Attached
+			if len(bgpNetworkName) != 0 {
+				bgpAttached = true
+			}
 		}
-		return constants.Unattached
 	}
 
-	node.Labels[constants.LabelUnderlayNetworkAttachment] = attachedToString(underlayAttached)
-	node.Labels[constants.LabelOverlayNetworkAttachment] = attachedToString(overlayAttached)
+	nodePatch := client.MergeFrom(node.DeepCopy())
+
+	updateAttachmentLabel := func(node *corev1.Node, key string, attached bool) {
+		if attached {
+			node.Labels[key] = constants.Attached
+			return
+		}
+		delete(node.Labels, key)
+	}
+
+	updateAttachmentLabel(node, constants.LabelUnderlayNetworkAttachment, underlayAttached)
+	updateAttachmentLabel(node, constants.LabelOverlayNetworkAttachment, overlayAttached)
+	updateAttachmentLabel(node, constants.LabelBGPNetworkAttachment, bgpAttached)
 
 	if err = r.Patch(ctx, node, nodePatch); err != nil {
 		log.Error(err, "unable to patch Node")
--- a/pkg/controllers/networking/pod_controller.go
+++ b/pkg/controllers/networking/pod_controller.go
@@ -57,8 +57,9 @@ const (
 )
 
 const (
-	IndexerFieldNode = "node"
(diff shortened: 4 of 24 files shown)


pkg/apis/networking/v1/types.go
```diff
@@ -19,16 +19,18 @@ package v1
 type NetworkType string
 
 const (
-	NetworkTypeUnderlay = NetworkType("Underlay")
-	NetworkTypeOverlay  = NetworkType("Overlay")
+	NetworkTypeUnderlay  = NetworkType("Underlay")
+	NetworkTypeOverlay   = NetworkType("Overlay")
+	NetworkTypeGlobalBGP = NetworkType("GlobalBGP")
```
mars1024, line 24: no need for a new network type, just need one new mode

hhyasdf: Actually we do need one, because we don't have the capability to specify network node while creating pods.

pkg/controllers/utils/client.go
```diff
@@ -188,17 +203,19 @@ func FindOverlayNetworkNetID(client client.Reader) (*int32, error) {
 	return nil, fmt.Errorf("no overlay network found")
 }
 
-func DetectNetworkAttachmentOfNode(client client.Reader, node *corev1.Node) (underlayAttached, overlayAttached bool, err error) {
+func DetectNetworkAttachmentOfNode(client client.Reader, node *corev1.Node) (underlayAttached,
```
mars1024, line 206: How about creating another function `detectBGPNetworkAttachementOfNode`?
Also `FindUnderlayNetworkForNode` won't work well if one node has vlan and bgp networks at same time.

pkg/apis/networking/v1/types.go
mars1024: I found that networkModeGlobalBGP is used in daemon part, and networkTypeGlobalBGP is used in webhook, how about treating globalBGP as one kind of underlay type? And use a function named `IsGlobalBGPMode` instead of introducing a new network type?

pkg/apis/networking/v1/utils.go
```diff
@@ -67,11 +67,23 @@ func GetNetworkMode(networkObj *Network) NetworkMode {
...
+	case NetworkTypeGlobalBGP:
+		if len(networkObj.Spec.Mode) == 0 {
+			return NetworkModeGlobalBGP
+		}
 	}
 
 	return networkObj.Spec.Mode
 }
 
+func IsGlobalNetwork(networkObj *Network) bool {
```
mars1024, line 79: What is `global` meaning here? IMO, globalBGP is not completely same as overlay. Also the capacity validation in webhook is not only for overlay and globalBGP, I believe we can do same validation on underlay network. So how about removing these two functions?

pkg/controllers/networking/pod_controller.go
```diff
@@ -183,61 +184,57 @@ func (r *PodReconciler) selectNetwork(pod *corev1.Pod) (string, error) {
...
 
-		// fall back to find underlay network by label selector
-		var underlayNetworkName string
-		if underlayNetworkName, err = utils.FindUnderlayNetworkForNodeName(r, pod.Spec.NodeName); err != nil {
-			return "", fmt.Errorf("unable to find underlay network for node %s", pod.Spec.NodeName)
-		}
-		if len(underlayNetworkName) == 0 {
-			return "", fmt.Errorf("no underlay network match node %s", pod.Spec.NodeName)
-		}
-		if !r.matchNetworkTypeInManager(underlayNetworkName, types.Underlay) {
```
mars1024, line 203: `matchNetworkTypeInManager` is helpful here because network cache in IPAM is syncing asynchronously, and this will disallow ip allocation if network type mismatch, which happened on a quick pod creation after network type changing.

hhyasdf: It seems that changing network type is not supported (and will not be supported) in validating webhook.

pkg/controllers/utils/client.go
hhyasdf: If `FindUnderlayNetworkForNode` cannot work well, maybe creating another `FindBGPUnderlayNetworkForNode` is ok but not another `detectBGPNetworkAttachementOfNode`?

pkg/controllers/networking/pod_controller.go
mars1024: yes, network type is unchangeable now, but delete and recreate is equal to "change" in an asynchronous system, I believe it's better to keep this logic.

pkg/controllers/networking/pod_controller.go
```diff
@@ -183,61 +184,57 @@ func (r *PodReconciler) selectNetwork(pod *corev1.Pod) (string, error) {
...
+	case types.GlobalBGP:
+		// try to get global bgp network by special node name
+		globalBGPNetworkName, err := r.getNetworkByNodeNameIndexer(GlobalBGPNodeName)
+		if err != nil {
+			return "", fmt.Errorf("unable to get overlay network by node name indexer: %v", err)
 		}
-		if !r.matchNetworkTypeInManager(overlayNetworkName, types.Overlay) {
-			return "", fmt.Errorf("network %s does not match type %q in manager", overlayNetworkName, types.Overlay)
+
+		if len(globalBGPNetworkName) != 0 {
```
mars1024, line 216: can you judge if returned network name is empty inside `getNetworkByNodeNameIndexer`, then we can save another 3 lines.

pkg/controllers/networking/node_controller.go
```diff
@@ -69,12 +69,22 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	var underlayAttached, overlayAttached bool
-	if underlayAttached, overlayAttached, err = utils.DetectNetworkAttachmentOfNode(r, node); err != nil {
+	var underlayAttached, overlayAttached, bgpAttached bool
+	if underlayAttached, overlayAttached, bgpAttached, err = utils.DetectNetworkAttachmentOfNode(r, node); err != nil {
```
mars1024, line 73: I'm afraid I still want bgp attachment detection is an independent function.

pkg/controllers/networking/node_controller.go
```diff
@@ -69,12 +69,22 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 		return ctrl.Result{}, err
 	}
 
+	// if global bgp network not exist, do not patch bgp attachment label to node
+	if bgpAttached {
+		if globalBGPNetwork, err := utils.FindGlobalBGPNetwork(r); err != nil {
+			log.Error(err, "unable to detect global bgp network exist")
+			return ctrl.Result{}, err
+		} else if len(globalBGPNetwork) == 0 {
+			bgpAttached = false
```
mars1024, line 84: if global bgp network not exists, I believe that not a bgp attachment label should be patched on nodes.

pkg/controllers/utils/client.go
mars1024: yes, just what I want, an independent function for BGP network attachment detection

pkg/apis/networking/v1/utils.go
mars1024: how about renaming it `IsUniqueNetwork`? seems more proper

pkg/controllers/utils/client.go
mars1024: Also we should take "one node belong to vlan and bgp at same time" into account, this will make bgp attachment detection not accurate.

pkg/controllers/networking/pod_controller.go
hhyasdf: It might be worthless, because you always need to check if the network exist and return an error if it's not exist, whether you return network name is empty inside `getNetworkByNodeNameIndexer` or not.

pkg/controllers/utils/client.go
hhyasdf: Done.

pkg/apis/networking/v1/utils.go
hhyasdf: It's renamed to `IsGlobalUniqueNetwork`.

pkg/controllers/networking/pod_controller.go
hhyasdf: Done.

pkg/controllers/networking/node_controller.go
hhyasdf: As you wish.

pkg/controllers/networking/node_controller.go
hhyasdf: It already is.

pkg/controllers/networking/node_controller.go
```diff
@@ -86,6 +104,7 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	node.Labels[constants.LabelUnderlayNetworkAttachment] = attachedToString(underlayAttached)
 	node.Labels[constants.LabelOverlayNetworkAttachment] = attachedToString(overlayAttached)
+	node.Labels[constants.LabelBGPNetworkAttachment] = attachedToString(bgpAttached)
```
mars1024, line 107: is it better to patch `LabelBGPNetworkAttachment` only if global bgp network exist?

pkg/controllers/networking/node_controller.go
mars1024: I mean the bgp attachment label should not exist, but not be set to false.

pkg/controllers/networking/pod_controller.go
```diff
@@ -183,55 +184,68 @@ func (r *PodReconciler) selectNetwork(pod *corev1.Pod) (string, error) {
...
-		if underlayNetworkName, err = utils.FindUnderlayNetworkForNodeName(r, pod.Spec.NodeName); err != nil {
-			return "", fmt.Errorf("unable to find underlay network for node %s", pod.Spec.NodeName)
-		}
-		if len(underlayNetworkName) == 0 {
-			return "", fmt.Errorf("no underlay network match node %s", pod.Spec.NodeName)
-		}
-		if !r.matchNetworkTypeInManager(underlayNetworkName, types.Underlay) {
-			return "", fmt.Errorf("network %s does not match type %q in manager", underlayNetworkName, types.Underlay)
+		if len(underlayNetworkName) != 0 {
+			if !r.matchNetworkTypeInManager(underlayNetworkName, types.Underlay) {
```
mars1024, line 193: can you restore these lines to 
```
if len(underlayNetworkName) == 0 {
    ...
}
if !r.matchNetworkTypeInManager(underlayNetworkName, types.Underlay) {
   ...
}
```

IMO, the nesting of several `if` is not very clear.

pkg/controllers/networking/node_controller.go
hhyasdf: Sorry, I misunderstood that such labels will be cleaned if the `attached` flag is false.

But I'm wondering if a node should be patched with a valued "false" attachment label if it's not be attached with an overlay/underlay/bgp network, maybe not patching or removing the no-attached label is the right behavior.

pkg/controllers/networking/pod_controller.go
hhyasdf: Good idea. It's done.

pkg/controllers/networking/node_controller.go
mars1024: There is a little different here, the true/false of underlay and overlay attachment label is always meaningful(correct), but if global bgp network not exists, the true/false of bpg attachment label will be **always** false, even this node belong to a normal underlay bgp network.
So i think it's better to remove this label if global bpg network not exists.

hhyasdf: In another words, "bgp attachment true/false label" is also meaningful (for literal meaning) if a bgp underlay network is attached or not, whether global bgp network not exists. We don't want it because it just seems useless and redundant if global bgp network is not exist.

I mean, for implementation, if labels seems useless and redundant (though it does have a  literal meaning), maybe we should treat them equally, which means, remove them.

mars1024: agreed，let's remove all the labels if it's false, because only true is meaningful for us.

hhyasdf: Done.

mars1024 (approved):
/lgtm
Thanks for your patience!

Issue #248, opened 2022-05-09
Nodes get "empty" quota while the Network still have available addresses to allocate


hhyasdf:
Bug Report
---

Type: *bug report*

### What happened
An Underlay Network (still with enough ip addresses to allocate) has 17 nodes, 16 of which have the "networking.alibaba.com/address-quota: empty" label while one of them has "networking.alibaba.com/address-quota: noempty" label.

The "noempty" one is the last item of the network.status.nodeList.

### What you expected to happen
All of the nodes have the "networking.alibaba.com/address-quota: noempty" label.

### Environment
- hybridnet version: v0.4.2
- OS (e.g. `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Kubernetes version: 1.20
- Install tools:
- Others:

PR #249, opened 2022-05-10
bugfix on closure function and fix debug log level

mars1024:
Pull Request Description
---

1. bug  fix on closure function
2. fix debug log level to 1, so it will be visible if logging is at debug level

### Special notes for reviews


hhyasdf (approved):
/lgtm

hhyasdf:
fix #248

Issue #218, opened 2022-04-13
Optimization of IP allocation performance in large-scale clusters


hhyasdf:
Issue Description
---

Type: *feature request*

### Describe what feature you want
By reducing the number of single direct request to apiserver (without cache) in manager's pod reconciling, we can get a significant increase of IP allocation performance.

For now, **four** requests exist during one pod reconciling process. It's very possible to reduce to **one**, which can bring a four times theoretical optimization to IP allocation.

### Additional context

What we might need mostly:
1. Remove the IP allocated annotation on Pod, and get the latest Pod allocation info in local cache.
2. Redesign the IPInstance CR to remove the proccess of updating of status from IP allocation.

PR #256, opened 2022-05-11
ip allocation performance optimize

hhyasdf:
Pull Request Description
---

Change IPInstance CR design and use a pod ip cache to accelerate ip allocation.

For #218

### Special notes for reviews


mars1024 (approved):
/lgtm
Let's bump up a major version.

PR #271, opened 2022-05-30
make manager unable to work with old crds

hhyasdf:
Pull Request Description
---

Make manager unable to work with old crds to avoid wrong upgrading of IPInstance objects.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #278, opened 2022-06-15
[CHERRY PICK] fix address duplication error while active-standby switch happens

hhyasdf:
…se-address-duplication

fix address duplication error while active-standby switch happens

Pull Request Description
---

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #280, opened 2022-06-17
add limitations for overlapped subnets

hhyasdf:
Pull Request Description
---

1. Subnets with different and overlapped CIDRs are not supported yet
2. Subnets with the same CIDRs but different ranges is supported, but only in the same Network
3. Rename some functions and add comments

### Special notes for reviews


mars1024 (approved):
/lgtm
/approve
Thanks!

PR #283, opened 2022-06-28
disable the automatic iptables mode detection of felix

hhyasdf:
Pull Request Description
---

Felix will use the automatic iptables mode detection of itself to use "iptables-nft-save/restore" or "iptables-legacy-save/restore", which is not stable enough. This pr remove the automatic iptables mode detection of felix and use [iptables-wrapper-installer.sh](https://github.com/alibaba/hybridnet/blob/main/dist/images/iptables-wrapper-installer.sh) to delete iptables mode of host.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #293, opened 2022-07-14
some dynamic integration tests and some bug fixes on failing cases

mars1024:
Pull Request Description
---

Append some Priority 1 and 2 integration tests, mainly about some dynamic checking after updating objects.
Also found and fixed some bugs through these cases,
1. allow terminating objects trigger network status reconcialization
2. use Node as core model in quota controller
3. trigger network and subnet status update after refreshing IPAM manager

### Special notes for reviews


hhyasdf (approved):
/lgtm

Issue #301, opened 2022-08-04
most pods can still be created even if webhook inactive


mars1024:
Issue Description
---

Type: *feature request*

### Describe what feature you want
Now, if webhook is down, all pods will be blocked from creation. But actually, most pods can be handled by controller even webhook inactive.

PR #302, opened 2022-08-04
unexpected pods could be handled in controller if webhook inactive

mars1024:
Pull Request Description
---

Sometime we do not want an inactive webhook will block all pods' creation, especially most pods can be handled by pod controller normally.
So we need some extended checks in controller to handle some unexpected pods if webhook inactive and notice our users.

#301

### Describe how you did it
1. mark pods if it went through webhook
2. pod controller will do some extra validations on unmarked pods
3. pod controller will check the network config consistency (network, network-type...) between controller and webhook on unmarked pods

### Describe how to verify it
NONE

### Special notes for reviews
NONE


hhyasdf (approved):
/lgtm

PR #305, opened 2022-08-11
make cni conf configurable

hhyasdf:
Pull Request Description
---

1. Rename the default name of cni conf file from `/etc/cni/net.d/00-hybridnet.conflist` to `/etc/cni/net.d/06-hybridnet.conflist`.
2. Remove the `bandwidth` plugin in cni conf.
3. Introduce some environment variables for `install-cni` container of daemon:
    - Use `CNI_CONF_NAME` to rename the cni conf file, e.g.,`04-hybridnet.conflist`.
    - Use `NEEDED_COMMUNITY_CNI_PLUGINS` to specify which cni community plugins need to be copied by hybridnet from inside container to the `/opt/cni/bin/` directory of host, e.g. `loopback,bandwidth`.
    -  Use `CNI_CONF_SRC` to specify which file in the container need to be copied to the `/etc/cni/net.d` directory of host, with which users can mount a custom cni conf file (e.g. using configmap as volume) into `install-cni` container and distribute it to every node.

### Special notes for reviews


mars1024 (commented):
a small question: why conflist file should be renamed?

hhyasdf:
> a small question: why conflist file should be renamed?

Because Kubernetes CNI runtime uses the alphabetically first file in the `/etc/cni/net.d/` directory.

mars1024 (approved):
/lgtm

PR #306, opened 2022-08-12
ensure v4 virt gateway reachable

hhyasdf:
Pull Request Description
---

`proxy_arp` will not work if the gateway address `169.251.1.1` is unreachable in route tables. Add `169.254.1.1 dev lo scope link` route if it happens.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #314, opened 2022-08-23
donot patch quota selector to ip retained pods

hhyasdf:
Pull Request Description
---

This fix the problem that if a stateful pod uses the last ip instance in underlay network it will always be Pending after it is deleted and automatically recreated. 

### Special notes for reviews


pkg/webhook/utils/utils.go
```diff
@@ -144,6 +144,14 @@ func ParseNetworkConfigOfPodByPriority(ctx context.Context, c client.Reader, pod
 				client.MatchingLabels{
 					constants.LabelPod: pod.Name,
 				})
+			if err != nil {
+				err = fmt.Errorf("parse pod %v/%v network config by exist ip instnaces: %v", pod.Namespace, pod.Name, err)
```
mars1024, line 148: small typo: instnaces --> instances

hhyasdf: fixed

mars1024 (approved):
/lgtm

PR #317, opened 2022-08-30
generate client-go lib

hhyasdf:
Pull Request Description
---

Use code-generator to generate clienset, informer and lister for export in `pkg/client`.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #326, opened 2022-10-19
modify ipv6 route cache gc parameters to adapt to kernel route error

hhyasdf:
Pull Request Description
---

IPv6 traffic's being dropped happens suddenly in some kernel versions (e.g., 4.18.0-80.el8.x86_64 of CentOS 8), while running "ip route get" for some of the ipv6 routes in table 39999 you can get a "Network is unreachable" error (though you can see a obviously correct route table configuration by running "ip route show"), and all neighbors are invalidated at the same time. This problem will shutdown all the Pods' network on the same node.

We believed that this problem is related to the kernel GC mechanism of ipv6 route cache because errors disappeared when the "net.ipv6.route.max_size" kernel parameter was configured to a much larger one (default 4096). But no related kernel patch is founded.

### Special notes for reviews

### DIFF
--- a/pkg/constants/sysctl.go
+++ b/pkg/constants/sysctl.go
@@ -46,4 +46,7 @@ const (
 
 	IPv6DisableModuleParameter = "/sys/module/ipv6/parameters/disable"
 	IPv6DisableSysctl          = "/proc/sys/net/ipv6/conf/%s/disable_ipv6"
+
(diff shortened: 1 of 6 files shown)


hhyasdf:
Refer to https://lore.kernel.org/netdev/20200305081747.tullbdlj66yf3w2w@mew.swordarmor.fr/

mars1024 (approved):
/lgtm

PR #327, opened 2022-10-22
ensure felix ready before start pod

hhyasdf:
Pull Request Description
---

Using felix as the NetworkPolicy implementation, pod will always be in `ContainerCreating` status if it has networking operations in its PostStart lifecycle hook.

> At that moment, pod is not even pingable from the pod's local node.

The root problem is that Kubernetes isn't updating the `podIP` field while the container lifecycle hooks are firing. Without this felix doesn't learn the pod's IP address and can't properly set up policy. Currently this seems to only be set after the pod moves out of ContainerCreating phase. Refer to https://github.com/projectcalico/calico/issues/3499.

This PR can resolve the problem properly.

NONE

### Describe how you did it
1. Introduce a `--patch-calico-pod-ips-annotation` flag for hybridnet daemon. If it's enabled, daemon will patch a `cni.projectcalico.org/podIPs` annotation to make felix be aware of what the pod's ip addresses are before initializing the pod's netns.
2. Introduce a `--check-pod-connectivity-from-host` flag for daemon. If it's enabled, daemon will check if the pod is pingable  before finishing initializing the pod's netns and fail if the pod is unpingable.

> PS: If the `--check-pod-connectivity-from-host` flag is enabled but `--patch-calico-pod-ips-annotation` is not, in a hybridnet-felix environment, pod will always be `ContainerCreating`.

### Describe how to verify it
To create a pod like:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ping-test2
spec:
  selector:
    matchLabels:
      app: ping-test2
  replicas: 1
  template:
    metadata:
      labels:
        app: ping-test2
    spec:
      containers:
      - name: ping
        image: hybridnet:v0.6.0
        command: ['bash', '-c', 'ping 8.8.8.8']
        lifecycle:
          postStart:
            exec:
              command:
              - bash
              - -c
              - 'ping -c 1 8.8.8.8 -W 600'
```

### Special notes for reviews

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0
 	github.com/emicklei/go-restful v2.16.0+incompatible
 	github.com/go-logr/logr v1.2.3
+	github.com/go-ping/ping v1.1.0
 	github.com/gogf/gf v1.16.6
 	github.com/heptiolabs/healthcheck v0.0.0-20211123025425-613501dd5deb
 	github.com/mdlayher/ethernet v0.0.0-20190606142754-0394541c37b7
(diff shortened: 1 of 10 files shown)


mars1024 (approved):
/lgtm

PR #331, opened 2022-11-03
introduce NodeInfo to replace Node annotations in daemon

hhyasdf:
Pull Request Description
---

Hybridnet-daemon pods need to collect and sync the overlay VTEP information to each others, now it's implemented by annotating Node objects. Considering list/watch Node objects on every node will cause a large bandwidth cost because the size of Node object is too large, we need to replace annotating Node objects by introduce a new CRD "NodeInfo".

NONE

### Describe how you did it
1. Introduce NodeInfo CRD to store the information which was annotated on Node objects
2. We need to make sure the overlay traffic will not be broken while updating hybridnet-daemon to the latest version.

### Special notes for reviews


pkg/controllers/multicluster/remotevtep_controller.go
```diff
@@ -76,21 +76,23 @@ func (r *RemoteVtepReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-	if !node.DeletionTimestamp.IsZero() {
+	if !nodeInfo.DeletionTimestamp.IsZero() {
 		log.V(1).Info("ignore terminating node")
 		_ = r.cleanVTEPForNode(ctx, req.Name)
 		return ctrl.Result{}, nil
 	}
 
-	var vtepIP, vtepMac, vtepVxlanIPList = node.Annotations[constants.AnnotationNodeVtepIP], node.Annotations[constants.AnnotationNodeVtepMac], node.Annotations[constants.AnnotationNodeLocalVxlanIPList]
+	var vtepIP, vtepMac, vtepVxlanIPList = nodeInfo.Spec.VTEPInfo.IP, nodeInfo.Spec.VTEPInfo.MAC,
```
mars1024, line 93: Should `nodeInfo.Spec.VTEPInfo` be checked whether to be nil here?

pkg/apis/networking/v1/types.go
```diff
@@ -108,4 +108,16 @@ type BGPPeer struct {
 	Password string `json:"password,omitempty"`
 }
 
+type VTEPInfo struct {
```
mars1024, line 111: This struct is both used by `networking/v1` and `multicluster/v1`, so is it better to place it on `pkg/apis/common/types.go`?

pkg/daemon/controller/controller.go
```diff
@@ -338,17 +338,17 @@ func (c *CtrlHub) handleVxlanInterfaceNeighEvent() error {
...
+				nodeInfo := &networkingv1.NodeInfo{}
 				nodeName := ipInstance.Labels[constants.LabelNode]
 
-				if err := c.mgr.GetClient().Get(context.TODO(), types.NamespacedName{Name: nodeName}, node); err != nil {
+				if err := c.mgr.GetClient().Get(context.TODO(), types.NamespacedName{Name: nodeName}, nodeInfo); err != nil {
 					return fmt.Errorf("failed to get node %v: %v", nodeName, err)
 				}
 
-				vtepMac, err = net.ParseMAC(node.Annotations[constants.AnnotationNodeVtepMac])
+				vtepMac, err = net.ParseMAC(nodeInfo.Spec.VTEPInfo.MAC)
```
mars1024, line 348: ditto, nil pointer check

pkg/daemon/controller/node_ip_cache.go
```diff
@@ -38,33 +36,30 @@ func NewNodeIPCache() *NodeIPCache {
...
+		if nodeInfo.Name == localNodeName {
 			continue
 		}
 
 		// ignore empty information
-		if _, exist := node.Annotations[constants.AnnotationNodeVtepMac]; !exist {
-			continue
-		}
-		if _, exist := node.Annotations[constants.AnnotationNodeLocalVxlanIPList]; !exist {
+		if len(nodeInfo.Spec.VTEPInfo.IP) == 0 ||
```
mars1024, line 52: ditto, nil pointer check

pkg/daemon/controller/nodeinfo.go
```diff
@@ -324,19 +324,50 @@ func ensureVxlanInterfaceAddresses(vxlanDev *vxlan.Device, addresses []netlink.A
...
+
+	_, err = controllerutil.CreateOrPatch(context.TODO(), r, nodeInfo, func() error {
+		if !nodeInfo.DeletionTimestamp.IsZero() {
+			return fmt.Errorf("node info %s is deleting, can not be updated", nodeInfo.Name)
+		}
+
+		nodeInfo.Spec.VTEPInfo = &networkingv1.VTEPInfo{
+			IP:       vtepIP.String(),
+			MAC:      vtepMac.String(),
+			LocalIPs: utils.GenerateIPStringList(nodeLocalVxlanAddr),
```
mars1024, line 344: sorting here is necessary because unordered list maybe fail in deep equal

pkg/daemon/controller/nodeinfo.go
```diff
@@ -324,19 +324,50 @@ func ensureVxlanInterfaceAddresses(vxlanDev *vxlan.Device, addresses []netlink.A
...
+		}
+
+		nodeInfo.OwnerReferences = []metav1.OwnerReference{
```
mars1024, line 347: `OwnerReferences` has no need to be assigned each time, if it already exists, just skip this step

pkg/daemon/controller/nodeinfo.go
```diff
@@ -324,19 +324,50 @@ func ensureVxlanInterfaceAddresses(vxlanDev *vxlan.Device, addresses []netlink.A
...
+		}
 
-	if oldNode.Annotations[constants.AnnotationNodeVtepIP] != newNode.Annotations[constants.AnnotationNodeVtepIP] ||
-		oldNode.Annotations[constants.AnnotationNodeVtepMac] != newNode.Annotations[constants.AnnotationNodeVtepMac] ||
-		oldNode.Annotations[constants.AnnotationNodeLocalVxlanIPList] != newNode.Annotations[constants.AnnotationNodeLocalVxlanIPList] {
-		return true
+		return nil
+	})
+
+	var updateTimestamp string
```
mars1024, line 355: if `CreateOrPatch` return `OperationNone` (aka, no-op), then timestamp should not be updated

pkg/daemon/controller/nodeinfo.go
```diff
@@ -345,22 +376,20 @@ func (r *nodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
 	if err := nodeController.Watch(&source.Kind{Type: &networkingv1.Network{}},
-		&fixedKeyHandler{key: ActionReconcileNode},
+		&fixedKeyHandler{key: ActionReconcileNodeInfo},
 		predicate.Funcs{
 			UpdateFunc: func(updateEvent event.UpdateEvent) bool {
-				return false
+				oldNetwork := updateEvent.ObjectOld.(*networkingv1.Network)
+				newNetwork := updateEvent.ObjectNew.(*networkingv1.Network)
+				return !utils2.StringSliceEqual(oldNetwork.Status.NodeList, newNetwork.Status.NodeList)
```
mars1024, line 392: if `network.Status.NodeList` changed, we believe `NodeInfo` will trigger a valid event, so here should always return false.

pkg/daemon/controller/nodeinfo.go
```diff
@@ -384,15 +413,15 @@ func (r *nodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	if feature.MultiClusterEnabled() {
 		if err := nodeController.Watch(&source.Kind{Type: &multiclusterv1.RemoteVtep{}},
-			&fixedKeyHandler{key: ActionReconcileNode},
+			&fixedKeyHandler{key: ActionReconcileNodeInfo},
 			predicate.Funcs{
 				UpdateFunc: func(updateEvent event.UpdateEvent) bool {
 					oldRemoteVtep := updateEvent.ObjectOld.(*multiclusterv1.RemoteVtep)
 					newRemoteVtep := updateEvent.ObjectNew.(*multiclusterv1.RemoteVtep)
 
 					if oldRemoteVtep.Spec.VTEPInfo.IP != newRemoteVtep.Spec.VTEPInfo.IP ||
```
mars1024, line 422: ditto, nil pointer check

pkg/utils/slice.go
```diff
@@ -44,6 +46,28 @@ func DeepEqualStringSlice(a []string, b []string) bool {
 	return true
 }
 
+func StringSliceEqual(slice1, slice2 []string) bool {
```
mars1024, line 49: why not just use `DeepEqualStringSlice` above?

pkg/controllers/multicluster/remotevtep_controller.go
hhyasdf: The VTEPInfo field for RemoteVtep object is inline and required, so it might be unnecessary.

pkg/daemon/controller/nodeinfo.go
hhyasdf: This is for the case that `NodeInfo` is deleted manually first and then `Node` be deleted. In this case, the event triggered by `NodeInfo` will not cause the delete operation of vxlan fdb entries on other nodes. If the next `Node` deletion step will not cause an event either, vxlan fdb entries on other nodes will not be cleaned correctly.

pkg/controllers/multicluster/remotevtep_controller.go
mars1024: Yes, in `RemoteVtep`, it is value type, but in `NodeInfo`, it is pointer type.

pkg/controllers/networking/ipam_controller_test.go
```diff
@@ -23,8 +23,8 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	ipammanager "github.com/alibaba/hybridnet/pkg/ipam/manager"
```
mars1024, line 26: seems unnecessary change, right?

pkg/daemon/controller/nodeinfo.go
hhyasdf: The VTEPInfo field of RemoteVtep is not a pointer.

pkg/daemon/controller/controller.go
hhyasdf: fixed

pkg/daemon/controller/node_ip_cache.go
hhyasdf: fixed

pkg/daemon/controller/nodeinfo.go
hhyasdf: fixed

pkg/daemon/controller/nodeinfo.go
hhyasdf: fixed

pkg/apis/networking/v1/types.go
hhyasdf: Let's do it in another PR.

pkg/controllers/multicluster/remotevtep_controller.go
hhyasdf: done

mars1024 (approved):
/lgtm

PR #333, opened 2022-11-07
introduce typha for large scale

hhyasdf:
Pull Request Description
---

1. Use the Calico Typha daemon to increase scale and reduce impact on the datastore.
2. Add init container to clean felix iptables rule automatically while policy is disabled.

### Special notes for reviews

### DIFF
--- a/charts/hybridnet/templates/deployments.yaml
+++ b/charts/hybridnet/templates/deployments.yaml
@@ -123,3 +123,109 @@ spec:
           ports:
             - containerPort: 9898
               name: webhook-port
+
+{{ if and .Values.typha .Values.daemon.enableFelixPolicy }}
+---
+# This manifest creates a Deployment of Typha to back the above service.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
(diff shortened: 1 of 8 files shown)


charts/hybridnet/templates/deployments.yaml
```diff
@@ -123,3 +123,109 @@ spec:
           ports:
             - containerPort: 9898
               name: webhook-port
+
+{{ if and .Values.typha .Values.daemon.enableFelixPolicy }}
```
mars1024, line 127: `.Values.typha` is struct, so should `.Values.typha.enabled` be introduced?

hhyasdf: I think maybe `--set typha=null` is enough for this, because `typha` struct should always to be checked in case it's empty.

mars1024 (approved):
/lgtm

mars1024:
ping @sjtufl , do you mind reviewing and merging this?

sjtufl (approved):
/lgtm

PR #343, opened 2022-11-24
fix: local externalTrafficPolicy access error

### LINKED ISSUE #341: Cannot connect to nodeport from inside the cluster while externalTrafficPolicy is Local
Bug Report
---

Type: *bug report*

### What happened
For a nodeport svc (endpoints are overlay pods), the externalTrafficPolicy of which is `Local`, the svc port of node (endpoints does exist on the node) cannot be accessed from other nodes inside the cluster (always timeout).

This kind of traffic should be successed.

### Environment
- hybridnet version: v0.7.5
- OS (e.g. `cat /etc/os-release`): CentOS 8
- Kernel (e.g. `uname -a`):
- Kubernetes version: 1.20
- Install tools:
- Others:

hhyasdf:
Pull Request Description
---

Fix #341 

### Describe how you did it
Add iptables rules to bypass marking the traffic of "local externalTrafficPolicy".

### Special notes for reviews

### DIFF
--- a/pkg/daemon/iptables/iptables.go
+++ b/pkg/daemon/iptables/iptables.go
@@ -45,7 +45,8 @@ const (
 	ChainHybridnetForward     = CustomChainPrefix + "FORWARD"
 	ChainHybridnetPreRouting  = CustomChainPrefix + "PREROUTING"
 
-	ChainHybridnetFromRuleSkip = CustomChainPrefix + "FROM-RULE-SKIP"
+	ChainHybridnetFromRuleSkip         = CustomChainPrefix + "FROM-RULE-SKIP"
+	ChainHybridnetPodToNodeTrafficMark = CustomChainPrefix + "POD-TO-NODE-MARK"
 
 	HybridnetOverlayNetSetName  = "HYBRIDNET-OVERLAY-NET"
 	HybridnetAllIPSetName       = "HYBRIDNET-ALL"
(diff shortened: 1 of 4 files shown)


pkg/daemon/iptables/iptables.go
```diff
@@ -428,16 +441,33 @@ func generateVxlanFilterRuleSpec(vxlanIf, allIPSet string, protocol Protocol) []
...
+
+func generatePodToNodeMarkRuleSpec() []string {
+	return []string{"-A", ChainHybridnetPodToNodeTrafficMark, "-m", "comment", "--comment", `"do pod -> node traffic mark"`,
 		"-j", "MARK", "--set-xmark", fmt.Sprintf("%s/%s", PodToNodeBackTrafficMarkString, PodToNodeBackTrafficMarkString),
 	}
 }
 
+// if pod -> node traffic is for svc with "externalTrafficPolicy: Local" do not mark it as overlay
+// TODO: this need to be retained if we use node ip routes to build the pod -> node traffic
+func generateLocalDNATedSkipRuleSpec(localNodeIP net.IP) []string {
```
mars1024, line 464: why not use ipset for local node IPs? maybe we can squash several rules into one

hhyasdf: Because ipset can only be used to match destination or source for a packet, it cannot be used together with `--ctorigdst` function.

mars1024 (approved):
/lgtm

PR #353, opened 2022-12-28
handle terminating stateful or vm pods with no owner reference

sjtufl:
Pull Request Description
---

It is possible that stateful (or vm) pod's owner reference is removed by generic garbage collection controller during fore-terminating phase, and we ought to allow pod controller to handle such kind of pods as long as they got ip-allocated finalizer.

NONE

### Describe how you did it
When handling terminating pods with no owner reference, fall back to looking up owner reference on corresponding IPInstances. If still got no owner reference, just remove pod's finalizer (which should be safe) and end this reconciliation.

### Special notes for reviews

### DIFF
--- a/pkg/controllers/networking/pod_controller.go
+++ b/pkg/controllers/networking/pod_controller.go
@@ -111,7 +111,36 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result
 	}
 
 	if pod.DeletionTimestamp != nil {
-		if strategy.OwnByStatefulWorkload(pod) {
+		var ownedObj client.Object = pod
+
+		// For terminating pods with no controller owner reference, try to get
+		// owner reference from ip instance.
+		if metav1.GetControllerOf(pod) == nil {
+			var ipInstanceList = &networkingv1.IPInstanceList{}
+			if err = r.List(ctx, ipInstanceList,
+				client.MatchingLabels{constants.LabelPod: pod.Name},
+				client.InNamespace(pod.Namespace),
+			); err != nil {
+				return ctrl.Result{}, wrapError("failed to list ip instance for pod", err)
+			}
+			for i := range ipInstanceList.Items {
+				ipInstance := &ipInstanceList.Items[i]
+				if !ipInstance.DeletionTimestamp.IsZero() {
+					continue
+				}
+				if metav1.GetControllerOf(ipInstance) != nil {
+					ownedObj = ipInstance.DeepCopy()
+					break
+				}
+			}
+
+			// If we still cannot find owner ref on all related ip instances, just remove pod finalizer.
+			if metav1.GetControllerOf(ownedObj) == nil {
+				return ctrl.Result{}, wrapError("unable to remove finalizer", r.removeFinalizer(ctx, pod))
+			}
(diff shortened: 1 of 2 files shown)


pkg/controllers/networking/pod_controller.go
```diff
@@ -120,7 +148,7 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result
 
 		if feature.VMIPRetainEnabled() {
 			// TODO: use APIReader to get VM/VMI object, because watch v1.VirtualMachine and v1.VirtualMachineInstance will always get errors
-			if isVMPod, vmName, _, err := strategy.OwnByVirtualMachine(ctx, pod, r.APIReader); isVMPod {
+			if isVMPod, vmName, _, err := strategy.OwnByVirtualMachine(ctx, ownedObj, r.APIReader); isVMPod {
```
hhyasdf, line 151: `strategy.OwnByVirtualMachine` should not be used to IPInstance, because IPInstance will be owned by VM but Pod will be owned by VMI (and VMI is owned by VM).

pkg/controllers/networking/pod_controller.go
```diff
@@ -739,10 +767,17 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) (err error) {
...
+					// phase, so we allow handling such kind of pods as long as
+					// they got ip allocated finalizer.
+					withNoControllerOwnerRef := metav1.GetControllerOf(pod) == nil &&
+						controllerutil.ContainsFinalizer(pod, constants.FinalizerIPAllocated)
+
 					ownByVMI, _ := strategy.OwnByVirtualMachineInstance(pod)
 
 					// terminating pods owned by stateful workloads and VMs should be processed for IP reservation
-					return strategy.OwnByStatefulWorkload(pod) || ownByVMI
+					return strategy.OwnByStatefulWorkload(pod) || ownByVMI || withNoControllerOwnerRef
```
mars1024, line 780: could you use lazy-judgment here to save some CPU?

pkg/controllers/networking/pod_controller.go
sjtufl: Refactored `strategy.OwnByVirtualMachine` a bit to handle directly and indirectly owned relationship.

pkg/controllers/networking/pod_controller.go
```diff
@@ -111,7 +111,35 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result
...
+					continue
+				}
+				if metav1.GetControllerOf(ipInstance) != nil {
+					ownedObj = ipInstance.DeepCopy()
+					break
+				}
+			}
+		}
+		// If we cannot find owner ref on pod nor ip instance, just remove pod finalizer.
+		if metav1.GetControllerOf(ownedObj) == nil {
```
mars1024, line 138: this judgement can only be false when `metav1.GetControllerOf(pod) == nil`, so these three lines should be moved inside the judgement section above.

hhyasdf (approved):
/lgtm

mars1024 (approved):
/lgtm

PR #358, opened 2023-01-31
chart: remove useless volume and add nodeSelector parameters

hhyasdf:
Pull Request Description
---

### Describe what this PR does / why we need it
1. Remove the useless hostPath `/var/run/` volume.
2. Add nodeSelector parameters in chart to make them configurable.
3. Release v0.5.11 chart.

### Special notes for reviews

PR #361, opened 2023-02-08
chart: mount /var/run/netns for containerd

hhyasdf:
Pull Request Description
---

`/var/run` hostPath volume was removed in #358 , this breaks pod's creation in containerd evironments. We need `/var/run/netns` to find netns for containerd Pods.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #366, opened 2023-02-21
fix: ipv6 subnet capacity calculation error because of int64 overflow

### LINKED ISSUE #357: Int64 will cause OutOfRange
https://github.com/alibaba/hybridnet/blob/34d622bacee76aab83b2ebdf7acd923fc0aa5175/pkg/utils/cidr.go#L76

Ipv6 address is 128bits, but 64bits of Int64, considering this circumstances
```
a := net.ParseIP("2001:db8:acad:3::101")
b := net.ParseIP("2001:db8:acad:3:ffff:ffff:ffff:fffe")
out := Capacity(a, b) // out is -258
```

then in subnet webhook, validating will pass. An over ranged subnet will be created, and causing OOM when  loading IPs of subnet into ipslce by the following code
```
//  /pkg/ipam/types/subnet.go Line 286

	// generate valid Available IP Slice
	s.AvailableIPs = NewIPSlice()
	for i := s.Start; utils.Cmp(i, s.End) <= 0; i = utils.NextIP(i) {
		if !s.Contains(i) {
			continue
		}
		// ignore reserved ip
		if s.IsReservedIP(i.String()) {
			continue
		}
		s.AvailableIPs.Add(i.String(), i.Equal(s.LastAllocatedIP))
	}
```

hhyasdf:
Pull Request Description
---

To correct the calculation of ipv6 subnet capacity.

Fix #357 

### Special notes for reviews

### DIFF
--- a/pkg/apis/networking/v1/utils.go
+++ b/pkg/apis/networking/v1/utils.go
@@ -19,6 +19,7 @@ package v1
 import (
 	"fmt"
 	"math"
+	"math/big"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


mars1024 (approved):
/lgtm

PR #367, opened 2023-02-22
improve: check accept ra sysctl before set v6 forward

hhyasdf:
Pull Request Description
---

If the ipv6 default route on host is generated by proto ra (by SLAAC), `accept_ra` sysctl need to be 2 if `ipv6 forwarding` is enable, or the default route might disappear after a while.

None.

### Special notes for reviews


pkg/daemon/utils/network.go
```diff
@@ -321,6 +321,28 @@ func EnableIPForward(family int) error {
 	if family == netlink.FAMILY_V4 {
 		return ip.EnableIP4Forward()
 	}
+
+	v6DefaultRoute, err := GetDefaultRoute(netlink.FAMILY_V6)
```
mars1024, line 325: how about checking and modifying accept_ra in `EnableIP6Forward?`

hhyasdf: `EnableIP6Forward` is in a third-party module, which cannot be modified.

mars1024 (approved):
/lgtm

PR #370, opened 2023-03-07
Support multiple bgp peers

hhyasdf:
Pull Request Description
---

1. Update gobgp to v3.11.0 and golang to 1.9.
2. Support add multiple bgp peers (bgp ToR is still needed, number of which need to be always one). Introduce two flags for items of `spec.config.bgpPeers`:
    1. `doesNotRouteTraffic`. If `true`, hybridnet will not add default route (next hop is this peer) for egress traffic of bgp pods.
    2. `allowNotEstablished`. If `true`, pod in the bgp network can be created if this peer is down.  

NONE.

### Special notes for reviews


mars1024 (approved):
lgtm

PR #378, opened 2023-04-26
improve: inherit node default gateway if no routable bgp peer exist

hhyasdf:
Pull Request Description
---

### Special notes for reviews

### DIFF
--- a/pkg/daemon/controller/utils.go
+++ b/pkg/daemon/controller/utils.go
@@ -270,7 +270,8 @@ func nodeBelongsToNetwork(nodeName string, network *networkingv1.Network) bool {
 }
 
 func collectGlobalNetworkInfoAndInit(ctx context.Context, client client.Reader, nodeVxlanIfName, nodeName string,
-	bgpManager *bgp.Manager, recordBGPPeers bool) (vxlanForwardNodeIfName string, bgpGatewayIP net.IP, err error) {
+	bgpManager *bgp.Manager, recordBGPPeers bool) (vxlanForwardNodeIfName string, attachedBGPNetworkExist bool,
+	bgpGatewayIP net.IP, err error) {
 
 	networkList := &networkingv1.NetworkList{}
 	if err = client.List(ctx, networkList); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/daemon/controller/utils.go
```diff
@@ -270,7 +270,8 @@ func nodeBelongsToNetwork(nodeName string, network *networkingv1.Network) bool {
 }
 
 func collectGlobalNetworkInfoAndInit(ctx context.Context, client client.Reader, nodeVxlanIfName, nodeName string,
-	bgpManager *bgp.Manager, recordBGPPeers bool) (vxlanForwardNodeIfName string, bgpGatewayIP net.IP, err error) {
+	bgpManager *bgp.Manager, recordBGPPeers bool) (vxlanForwardNodeIfName string, attachedBGPNetworkExist bool,
```
mars1024, line 273: the return "attachedBGPNetworkExist" may always be false

pkg/daemon/route/utils.go
```diff
@@ -402,19 +402,52 @@ func ensureRoutesForVlanSubnet(forwardLink netlink.Link, cidr *net.IPNet, gatewa
...
+	}, netlink.RT_FILTER_TABLE)
+	if err != nil {
+		return fmt.Errorf("failed to list route for table %v: %v", table, err)
+	}
+
+	for _, route := range routeList {
+		// cannot use route.Equal() because of empty fields
+		if daemonutils.IsDefaultRoute(&route, family) &&
+			// TODO: support multiple bgp gateway
+			(!route.Gw.Equal(defaultRoute.Gw) || route.LinkIndex != defaultRoute.LinkIndex) {
```
mars1024, line 444: can we restrict the deletion scope here, in other words, is the judgement for the route deletion safe enough?

pkg/daemon/controller/utils.go
hhyasdf: fixed

pkg/daemon/route/utils.go
hhyasdf: Yes, it is safe, only daefault routes with different gateway and interface will be deleted.

mars1024 (approved):
/lgtm

PR #385, opened 2023-06-15
Support pod name over 63 characters

### LINKED ISSUE #375: Pod name can't be more than 63 characters in length
Bug Report
---

Type: *bug report*

### What happened
If we want to create a pod, the name of which has a length more than 63 characters, it will be stuck in `ContainerCreating` status and cannot be `Running`. "unable to couple IPs" error happens.

This pod can be created successfully.

### Environment
- hybridnet version: v0.8.0
- OS (e.g. `cat /etc/os-release`): Anolis OS
- Kernel (e.g. `uname -a`): 4.19
- Kubernetes version: v1.22
- Install tools: sealer
- Others:

hhyasdf:
Pull Request Description
---

Make pods, the name of which has a length more than 63 characters, can be created successfully.

Fixes #375 

### Describe how you did it
If the length pod name is over 63 characters, hash it to a 63-characters string when it needs to be as a label value of IPInstance.

### Special notes for reviews

### DIFF
--- a/pkg/utils/transform/ipam.go
+++ b/pkg/utils/transform/ipam.go
@@ -17,6 +17,8 @@
 package transform
 
 import (
+	"crypto/md5"
+	"fmt"
 	"net"
 
 	v1 "github.com/alibaba/hybridnet/pkg/apis/networking/v1"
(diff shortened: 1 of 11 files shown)


pkg/utils/transform/ipam.go
```diff
@@ -77,6 +80,20 @@ func TransferIPInstancesForIPAM(ips []*v1.IPInstance) []*ipamtypes.IP {
 	return ret
 }
 
+func TransferPodNameForLabelValue(podName string) string {
+	if len(podName) == 0 {
```
mars1024, line 84: this judgement seems useless :)

mars1024 (dismissed):
/lgtm

pkg/utils/transform/ipam.go
hhyasdf: done~

mars1024 (approved):
/lgtm

Issue #383, opened 2023-06-13
The wrong interfaces in CNI result generated by hybridnet plugin


mars1024:
Bug Report
---

Type: *bug report*

### What happened

As CNI spec, the interfaces of CNI result should be the ones which were created inside container. Then it will be passed through CNI chain, and being operated by some other chained plugins (e.g. tuning, sbr...).

### What you expected to happen

CNI result generated by hybridnet plugin should contain the created interfaces (eth0 inside container).

### How to reproduce it (as minimally and precisely as possible)

NONE

### Anything else we need to know?

NONE

### Environment
- hybridnet version: latest
- OS (e.g. `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Kubernetes version:
- Install tools:
- Others:

PR #388, opened 2023-06-21
add created host interface into CNI result for chained plugins like bandwidth

mars1024:
Pull Request Description
---

Created host interface should also be placed into CNI result, because the chained bandwidth plugin will use this for traffic control.

#383 

### Describe how you did it
NONE

### Describe how to verify it
NONE

### Special notes for reviews
NONE


hhyasdf (approved):
/lgtm

PR #391, opened 2023-06-28
bugfix: simple trigger source nilptr and ipset length error

hhyasdf:
Pull Request Description
---

1. Fix nil pointer error which might happen in daemon during initializiation.
2. Fix the error that ipv6 "local underlay net"  ipset name is longer than 31 characters.  

### Special notes for reviews

### DIFF
--- a/pkg/daemon/iptables/iptables.go
+++ b/pkg/daemon/iptables/iptables.go
@@ -49,11 +49,14 @@ const (
 	ChainHybridnetFromRuleSkip         = CustomChainPrefix + "FROM-RULE-SKIP"
 	ChainHybridnetPodToNodeTrafficMark = CustomChainPrefix + "POD-TO-NODE-MARK"
 
-	HybridnetOverlayNetSetName       = "HYBRIDNET-OVERLAY-NET"
-	HybridnetAllIPSetName            = "HYBRIDNET-ALL"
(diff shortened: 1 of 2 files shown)


pkg/daemon/iptables/iptables.go
```diff
@@ -49,11 +49,14 @@ const (
...
 
-	HybridnetOverlayNetSetName       = "HYBRIDNET-OVERLAY-NET"
-	HybridnetAllIPSetName            = "HYBRIDNET-ALL"
-	HybridnetNodeIPSetName           = "HYBRIDNET-NODE-IP"
-	HybridnetLocalPodIPSetName       = "HYBRIDNET-LOCAL-POD-IP"
-	HybridnetLocalUnderlayNetSetName = "HYBRIDNET-LOCAL-UNDERLAY-NET"
+	// The origin ip set name below should not be longer than 25 characters, because v6 ip set name will get an "inet6:" prefix,
+	// and the actual length of ip set name should not be longer than 31 characters.
+
+	HybridnetOverlayNetSetName       = "HYBR-OVERLAY-NET"
```
mars1024, line 55: will the old ipsets still exist after upgrading?

hhyasdf: noop~

mars1024 (approved):
/lgtm

PR #402, opened 2023-09-20
feat: introduce update-ipinstance-status for daemon

hhyasdf:
Pull Request Description
---

So we can make daemon donot update status for ipinstance. And remove `status.phase` field for IPInstance, which is useless.

### Special notes for reviews


mars1024 (approved):
/lgtm

PR #403, opened 2023-09-21
only reserver ip instance after pod terminated

mars1024:
Pull Request Description
---

IPInstance reservation happens immediately after pod terminating now, so `preStop` will lose network connectivity because of this.

NONE

### Describe how you did it
To avoid losing network connectivity when pre-stopping, we should reserve ip instance only after pod terminated, which means all containers of pod do not need network connectivity any more.

### Describe how to verify it
Integration test will do this.

### Special notes for reviews
NONE


hhyasdf (approved):
/lgtm

PR #407, opened 2023-10-10
bugfix: waiting containers block pod termination

### LINKED ISSUE #406: Statefulset's pod got stuck in terminating forever after deleting
Bug Report
---

Type: *bug report*

### What happened

Statefulset's pod got stuck in terminating forever after deleting, and container state in the pod status shows as following

```
containerStatuses:
  image: ""
  lastState:
      terminated:
        exitCode: 137
        finishedAt: null
        message: The container could not be located when the pod was deleted.  The
          container used to be Running
        reason: ContainerStatusUnknown
        startedAt: null
  name: xxx-container
  ready: false
  restartCount: 0
  started: false
  state:
    waiting:
      reason: ContainerCreating
```

Pod can be deleted.

### Environment
- hybridnet version: v0.8.7
- Kernel (e.g. `uname -a`):
- Kubernetes version:
- Install tools:
- Others:

hhyasdf:
Pull Request Description
---

As mentioned in #406, now sts pods might get stuck in "Terminating" because of unknow _waiting_ container state. To check if pod is completely terminated, we should check if "all the containers' state are not _running_" rather than "all the containers‘ state are _terminated_".

Fixes #406 

### Special notes for reviews

### DIFF
--- a/pkg/controllers/utils/pod.go
+++ b/pkg/controllers/utils/pod.go
@@ -51,9 +51,10 @@ func PodIsScheduled(pod *v1.Pod) bool {
 	return len(pod.Spec.NodeName) > 0
 }
 
-func PodIsTerminated(pod *v1.Pod) bool {
+func PodIsNotRunning(pod *v1.Pod) bool {
+	// check if all the pod containers are not running, if any container is running, pod is still running
 	for i := range pod.Status.ContainerStatuses {
-		if pod.Status.ContainerStatuses[i].State.Terminated == nil {
+		if pod.Status.ContainerStatuses[i].State.Running != nil {
 			return false
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/controllers/utils/pod.go
```diff
@@ -52,8 +52,9 @@ func PodIsScheduled(pod *v1.Pod) bool {
 }
 
 func PodIsTerminated(pod *v1.Pod) bool {
```
sjtufl, line 54: What about changing func name to something like `PodHasNoRunningContainer`?

sjtufl (approved):
/lgtm

pkg/controllers/utils/pod.go
hhyasdf: done~

mars1024 (approved):
/lgtm

Issue #411, opened 2024-02-28
Skip IPAM for specific pods


PixelPixel00:
Issue Description

In my use case, we only need hybridnet-manager to manage address for some of the pods in cluster. I want to add some mechanisms (e.g. label selector) to skip managing adress for specific pods
---

Type: *feature request*

PR #412, opened 2024-03-01
implement pod selector and use it in predication of pod reconciliations

mars1024:
Pull Request Description
---

Support configurable pod label selector for selecting specified pods for IPAM.

#411 

### Describe how you did it
NONE

### Describe how to verify it
NONE

### Special notes for reviews

### DIFF
--- a/pkg/controllers/networking/pod_controller.go
+++ b/pkg/controllers/networking/pod_controller.go
@@ -80,6 +80,7 @@ type PodReconciler struct {
 	IPAMManager IPAMManager
 
 	concurrency.ControllerConcurrency
+	PodSelector utils.PodSelector
 }
 
 //+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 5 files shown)


pkg/controllers/networking/pod_controller.go
```diff
@@ -773,6 +774,19 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) (err error) {
...
+					if r.PodSelector == nil {
+						return true
+					}
+
+					pod, ok := obj.(*corev1.Pod)
+					if !ok {
+						return false
+					}
+
+					// Only selected pods should be processed
```
sjtufl, line 787: Except for Pod controller, should we also let  mutating & validating webhook be aware of the pod selector?

mars1024: yes, objectSelector field of `webhookConfigurations` should be set as same as this.

sjtufl (approved):
/lgtm

PR #418, opened 2024-05-27
Bugfix:  recycle unneeded reserved IPs for pre-assigned pods

### LINKED ISSUE #417: Changing specified IP of stateful workload will cause multiple IPs being reserved
Bug Report
---

Type: *bug report*

### What happened
1. Specify ip for StatefulSet pod by adding annotation `networking.alibaba.com/ip-pool: 100.64.99.111` (sts replica is  set to 1 for demo)
2. Change annotation value to `100.64.99.112`, new pod will be created with ip `100.64.99.112`, and the previously specified `100.64.99.111` is also reserved, which is unintended.
```
~ # kubectl get pods -owide | grep 100.64.99.112
statefulset-e7v2c-0                             1/1     Running   0          97s     100.64.99.112
~ # kubectl get ipinstance | grep statefulset
100-64-99-111   100.64.99.111/16             statefulset-e7v2c-0                                          init     init
100-64-99-112   100.64.99.112/16             statefulset-e7v2c-0                izbp149b3nd59ljk7lbea7z   init     init
```
3. If we delete pod now, multiple reserved IPs will cause logic disruptions on inference of ip family when doing stateful allocate for newly created pod.
```
unable to stateful allocate with provider [hybridnet]: unable to assign: must assign two IPs when DualStack family, but [{ 100.64.99.112}]
```

Do not reserve previously specified IPs.

### Environment
- hybridnet version: 
- Kernel (e.g. `uname -a`):
- Kubernetes version:
- Install tools:
- Others:

sjtufl:
Pull Request Description
---

Fixes #417 

### Describe how you did it
Recycle unneeded reserved IPs before assigning IPs to pods.

### Special notes for reviews

### DIFF
--- a/pkg/controllers/networking/pod_controller.go
+++ b/pkg/controllers/networking/pod_controller.go
@@ -27,6 +27,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


mars1024:
basically LGTM，but we need more test cases，do you mind adding some?

sjtufl:
> basically LGTM，but we need more test cases，do you mind adding some?

Added, PTAL.

mars1024 (approved):
/approve
Thanks!

